PR #1, opened 2019-09-27
Support for RevertSP in library

jakubradtke:
This patch implements library functions for REVERTSP OPAL method.
Only NVMe PT path supported as no kernel IOCTL exists for the method.
RevertSP functionality can be used in scenario where user data shall
not be erased after TPer reverting (please see "TCG Storage Security
Subsystem Class: Opal", Section "RevertSP - Base Template SP Method"
for more details).

Signed-off-by: Jakub Radtke <jakub.radtke@intel.com>

### DIFF
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -73,6 +73,11 @@ int sed_reverttper(struct sed_device *dev, const struct sed_key *key, bool psid)
 /**
  *
  */
+int sed_revertsp(struct sed_device *dev, const struct sed_key *key, bool keep_global_rn_key);
+
+/**
+ *
+ */
 int sed_setpw(struct sed_device *dev, const struct sed_key *old_key,
 		const struct sed_key *new_key);
 
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -74,7 +74,7 @@ int sed_reverttper(struct sed_device *dev, const struct sed_key *key, bool psid)
  * Revert Locking SP to its Original Factory State.
  *
  * @param dev			the device to operate on
- * @param key			the SID password
+ * @param key			the Admin1 password
  * @param keep_global_rn_key	if true then TPer shall continue to use the media
  * 				encryption key associated with the Global locking
  * 				range after the Locking SP transitions to the
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -71,9 +71,20 @@ int sed_lock_unlock(struct sed_device *dev, const struct sed_key *key, enum SED_
 int sed_reverttper(struct sed_device *dev, const struct sed_key *key, bool psid);
 
 /**
+ * Revert Locking SP to its Original Factory State.
  *
+ * @param dev			the device to operate on
+ * @param key			the SID password
+ * @param keep_global_rn_key	if true then TPer shall continue to use the media
+ * 				encryption key associated with the Global locking
+ * 				range after the Locking SP transitions to the
+ * 				“Manufactured-Inactive” state. Can be used in order
+ * 				to skip data erasing when OPAL functionality is
+ * 				reverted.
+ *
+ * @return OPAL_SUCCESS on success
  */
-int sed_revertsp(struct sed_device *dev, const struct sed_key *key, bool keep_global_rn_key);
+int sed_revertlsp(struct sed_device *dev, const struct sed_key *key, bool keep_global_rn_key);
 
 /**
  *
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -161,6 +161,10 @@ static uint8_t opal_method[][OPAL_UID_LENGTH] = {
 		{ 0x00, 0x00, 0x00, 0x06, 0x00, 0x00, 0x08, 0x03 },
 };
 
+static const uint8_t opal_mehod_params[][OPAL_UID_LENGTH] = {
+	[OPAL_REVERTSP_KEEP_GL_RN_KEY] = {0x83, 0x06, 0x00, 0x00},
+};
+
 struct opal_req_item {
 	int type;
 	int len;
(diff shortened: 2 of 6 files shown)


src/lib/include/libsed.h
```diff
@@ -70,6 +70,11 @@ int sed_lock_unlock(struct sed_device *dev, const struct sed_key *key, enum SED_
  */
 int sed_reverttper(struct sed_device *dev, const struct sed_key *key, bool psid);
 
+/**
+ *
+ */
+int sed_revertsp(struct sed_device *dev, const struct sed_key *key, bool keep_global_rn_key);
```
AndrzejJakowski: Suggest renaming function to sed_revertlsp to indicate that revert is happening on Locking SP.

src/lib/nvme_pt_ioctl.c
```diff
@@ -161,6 +161,10 @@ static uint8_t opal_method[][OPAL_UID_LENGTH] = {
 		{ 0x00, 0x00, 0x00, 0x06, 0x00, 0x00, 0x08, 0x03 },
 };
 
+static const uint8_t opal_mehod_params[][OPAL_UID_LENGTH] = {
+	[OPAL_REVERTSP_KEEP_GL_RN_KEY] = {0x83, 0x06, 0x00, 0x00},
```
AndrzejJakowski: KeepGlobalRangeKey parameter is 0x00600000 , 0x83 byte should by using opal parser API (see append_short_atom* functions family)

src/lib/nvme_pt_ioctl.c
```diff
@@ -1678,6 +1713,28 @@ int opal_activate_lsp_pt(struct sed_device *dev, const struct sed_key *key,
 	return ret;
 }
 
+int opal_revertsp_pt(struct sed_device *dev, const struct sed_key *key, bool keep_global_rn_key)
```
AndrzejJakowski: suggest changing function name  to indicate that Locking SP is reverted

src/lib/nvme_pt_ioctl.c
```diff
@@ -825,6 +829,37 @@ static int opal_revert_tper_local(int fd, struct opal_device *dev)
...
+	{.type = OPAL_U8, .len = 1,
+		.val = {.byte = OPAL_TRUE}},
+	{.type = OPAL_U8, .len = 1,
+		.val = {.byte = OPAL_ENDNAME}},
+};
+
+static int opal_revert_sp_local(int fd, struct opal_device *dev, bool keep_global_rn_key) {
+    int ret = 0;
+
+    prepare_req_buf(dev, opal_revert_sp_cmd, ARRAY_SIZE(opal_revert_sp_cmd),
```
AndrzejJakowski: please conform to the coding standard: use tabs for indentation instead of spaces

AndrzejJakowski (changes requested):

src/lib/include/libsed.h
jakubradtke: done

src/lib/nvme_pt_ioctl.c
jakubradtke: done

src/lib/nvme_pt_ioctl.c
jakubradtke: done

src/lib/nvme_pt_ioctl.c
jakubradtke: done

src/lib/include/libsed.h
```diff
@@ -70,6 +70,22 @@ int sed_lock_unlock(struct sed_device *dev, const struct sed_key *key, enum SED_
  */
 int sed_reverttper(struct sed_device *dev, const struct sed_key *key, bool psid);
 
+/**
+ * Revert Locking SP to its Original Factory State.
+ *
+ * @param dev			the device to operate on
+ * @param key			the SID password
```
AndrzejJakowski: It should be Admin1 password

src/lib/sed.c
```diff
@@ -72,6 +74,7 @@ static struct opal_interface opal_if = {
 	.ownership_fn = sedopal_takeownership,
 	.revert_fn = sedopal_reverttper,
 	.activatelsp_fn = sedopal_activatelsp,
+	.revertsp_fn = sedopal_revertsp,
```
AndrzejJakowski: Suggest to set it to NULL and handle like in other functions, e.g. 


int sed_ds_anybody_read(struct sed_device *dev, uint8_t *to, uint32_t size, uint32_t offset)
{
	if (curr_if->ds_anybody_read_fn == NULL)
		return -EOPNOTSUPP;

	return curr_if->ds_anybody_read_fn(dev, to, size, offset);
}

Then sed_ioctl.[ch] would not have to be changed

AndrzejJakowski (changes requested):

src/lib/sed.c
RevanthRajashekar: With this being added, I would suggest removing the corresponding function in the sed_ioctl.c and sed_ioctl.h (bcoz, we aren't even reaching there).

RevanthRajashekar (commented):
With the changes suggested, the patch looks fine for me. Are we planning on exposing this feature to the user..?

src/lib/include/libsed.h
jakubradtke: Yep, SID is not correct here as this is used for TPer Owner, initially it is the same as activate SP is copying SID to Admin1.

jakubradtke:
> With the changes suggested, the patch looks fine for me. Are we planning on exposing this feature to the user..?

In my opinion it is valuable to use this functionality in order to implement revert TPer w/o data erasing. Having separate option for revert LSP in sedcli ... hmm, might be not needed

src/lib/sed.c
jakubradtke: done

AndrzejJakowski:
> > With the changes suggested, the patch looks fine for me. Are we planning on exposing this feature to the user..?
> 
> In my opinion it is valuable to use this functionality in order to implement revert TPer w/o data erasing. Having separate option for revert LSP in sedcli ... hmm, might be not needed

Let's keep it in library for now and see if there is interest to have this functionality from sedcli users

PR #2, opened 2019-09-30
Replace hardcoded value of header for parameter in SUM activation method

AndrzejJakowski:
This patch replaces hardcoded header value for SingleUserModeSelectionList
parameter for SUM activation. Original code had short atom header with
value 0x83 added to it. Current code dynamically appends header
instead of hardcoding it.


jakubradtke (approved):

RevanthRajashekar (commented):
Looks good for me :)

PR #4, opened 2019-10-03
Tests: improvements in test framework

AndrzejJakowski:


test/test_sedcli_basic.py
```diff
@@ -0,0 +1,67 @@
...
+import os
+import sys
+import unittest
+import subprocess
+import random
+import string
+
+class TestSedcliBasic(unittest.TestCase):
+
+	NVME_DEV_PATH = '/dev/nvme0n1'
```
RevanthRajashekar, line 32: The device path may change,  need not have to be nvme0n1 every time.
I think it would be a better idea to ask the user to enter the device path

test/test_sedcli_basic.py
```diff
@@ -0,0 +1,67 @@
...
+		outs, errs = pipe.communicate()
+		pipe.stdin.close()
+
+		self.assertEqual(pipe.returncode, 0, "sedcli returned error while setting RLE and WLE on global range:\n" + str(errs))
+		self.assertEqual(len(errs), 0, "sedcli returned success code but error message has been printed on stderr:\n" + str(errs))
+
+# -----------------------------------------------------------------------------
+if __name__ == '__main__':
+	if len(sys.argv) > 1:
+		TestSedcliBasic.NVME_DEV_PATH = sys.argv.pop()
```
RevanthRajashekar, line 67: So I think this is where the device name is taken from the user

RevanthRajashekar:
The patch looks fine for me :)

test/test_sedcli_basic.py
jakubradtke: or maybe have a config file

test/test_sedcli_basic.py
```diff
@@ -0,0 +1,67 @@
...
+		Tests for: sedcli intial provision of drive (that is take ownerhisp,
+		activate Locking SP and setup global locking range)
+		"""
+
+		print("Key used to initial provision drive:" + str(self.nvme_sid_key))
+
+		pipe = subprocess.Popen(["sedcli", "--ownership", "--device", self.NVME_DEV_PATH], stdout = subprocess.PIPE, stdin = subprocess.PIPE, stderr = subprocess.PIPE)
+		pipe.stdin.write(self.nvme_sid_key + b'\n' + self.nvme_sid_key + b'\n')
+		outs, errs = pipe.communicate()
+		pipe.stdin.close()
```
jakubradtke: Utility function can be implemented in order to execute commands and get result. This will make the code more readable

jakubradtke (approved):
I have some comments but they are just improvements not issues

test/test_sedcli_basic.py
AndrzejJakowski: NVME_DEV_PATH = '/dev/nvme0n1' if the default setting that can be overriden when user specifies command line parameter. See below. I agree with you there are better ways to handle static configuration e.g. through config file

if __name__ == '__main__':
	if len(sys.argv) > 1:
		TestSedcliBasic.NVME_DEV_PATH = sys.argv.pop()

test/test_sedcli_basic.py
AndrzejJakowski: yep

AndrzejJakowski:
Ready for review after incorporating Jakub's feedback

PR #6, opened 2019-10-04
sedcli: Changing loglevel from LOG_ERROR to LOG_INFO

RevanthRajashekar:
to print the sed_err_to_text string

The existing log level LOG_ERROR prints sed_err_to_text string into stderr
By changing the log level to LOG_INFO, we now print the string into stdout

### DIFF
--- a/src/sedcli_main.c
+++ b/src/sedcli_main.c
@@ -329,10 +329,33 @@ static int get_password(struct sed_key *pwd)
 	return 0;
 }
 
+static void print_sed_status(int status)
+{
+	const char *sed_status = NULL;
+
+	if (status < 0) {
+		if (status == -EINVAL) {
+			sedcli_printf(LOG_ERR, "Invalid parameter\n");
+		} else if (status == -ENODEV) {
+			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
+		} else if (status == -ENOMEM) {
+			sedcli_printf(LOG_ERR, "No memory\n");
+		} else {
+			sedcli_printf(LOG_ERR, "Unknown error\n");
+		}
+	} else {
+		sed_status = sed_error_text(status);
+		if (sed_status == NULL)
+			sedcli_printf(LOG_ERR, "Unknown Error\n");
+		else
+			sedcli_printf((status == SED_SUCCESS) ? LOG_INFO : LOG_ERR,
+					"%s\n", sed_status);
+	}
+}
+
 static int handle_ownership(void)
 {
 	struct sed_device *dev = NULL;
-	const char *sed_status = NULL;
 	int ret;
 
 	sedcli_printf(LOG_INFO, "New SID password: ");
@@ -364,17 +387,7 @@ static int handle_ownership(void)
 
 	ret = sed_takeownership(dev, &opts->pwd);
 
-	sed_status = sed_error_text(ret);
-
-	if (sed_status == NULL && ret != 0) {
-		if (ret == -EINVAL) {
-			sedcli_printf(LOG_ERR, "Invalid parameter\n");
-		} else {
-			sedcli_printf(LOG_ERR, "Unknown error\n");
-		}
-	} else {
-		sedcli_printf(LOG_ERR, "%s\n", sed_status);
-	}
+	print_sed_status(ret);
 
 	sed_deinit(dev);
 
@@ -384,7 +397,6 @@ static int handle_ownership(void)
 static int handle_activatelsp(void)
 {
 	struct sed_device *dev = NULL;
-	const char *sed_status = NULL;
 	int ret;
 
 	sedcli_printf(LOG_INFO, "Enter SID password: ");
@@ -403,19 +415,7 @@ static int handle_activatelsp(void)
 
 	ret = sed_activatelsp(dev, &opts->pwd);
 
-	sed_status = sed_error_text(ret);
-
-	if (sed_status == NULL && ret != 0) {
-		if (ret == -EINVAL) {
-			sedcli_printf(LOG_ERR, "Invalid parameter\n");
-		} else if (ret == -ENODEV) {
-			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
-		} else {
-			sedcli_printf(LOG_ERR, "Unknown error\n");
-		}
-	} else {
-		sedcli_printf(LOG_ERR, "%s\n", sed_status);
-	}
+	print_sed_status(ret);
 
 	sed_deinit(dev);
 
@@ -425,7 +425,6 @@ static int handle_activatelsp(void)
 static int handle_reverttper(void)
 {
 	struct sed_device *dev = NULL;
-	const char *sed_status = NULL;
 	int ret;
 
 	sedcli_printf(LOG_INFO, "Enter %s password: ", opts->psid ? "PSID" : "SID");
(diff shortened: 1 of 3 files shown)


src/sedcli_main.c
```diff
@@ -593,7 +593,7 @@ static int handle_setpw(void)
 			sedcli_printf(LOG_ERR, "Unknown error\n");
```
jakubradtke: looks like this code is repeated multiple times. Maybe create separate function for this?

AndrzejJakowski: > looks like this code is repeated multiple times. Maybe create separate function for this?

Sounds like a good idea

RevanthRajashekar: Done

src/sedcli_main.c
```diff
@@ -329,10 +329,32 @@ static int get_password(struct sed_key *pwd)
...
+		} else if (status == -ENOMEM) {
+			sedcli_printf(LOG_ERR, "No memory\n");
+		} else {
+			sedcli_printf(LOG_ERR, "Unknown error\n");
+		}
+	 } else {
+		 sedcli_printf(LOG_INFO, "%s\n", sed_status);
+	 }
+
+	sed_deinit(dev);
```
AndrzejJakowski: sed_deinit doesn't logically fit in here, sed_status_ret function should be responsible for interpreting and printing an error and status, it should not be responsible for deinitialization of sed_device context.
Please remove sed_deinit and struct sed_device function argument because it doesn't belong here

RevanthRajashekar: Or does it make sense to add (_and_deinit) to the function name.
Anyway we do deinit in every handle functions

AndrzejJakowski: > Or does it make sense to add (_and_deinit) to the function name.
> Anyway we do deinit in every handle functions

I don't think it make sense. What if somebody decides to deallocate device context by himself and just print the message. Your function would require him to provide dev context...

RevanthRajashekar: ok :)

RevanthRajashekar: done

src/sedcli_main.c
```diff
@@ -329,10 +329,30 @@ static int get_password(struct sed_key *pwd)
...
+		if (status == -EINVAL) {
+			sedcli_printf(LOG_ERR, "Invalid parameter\n");
+		} else if (status == -ENODEV) {
+			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
+		} else if (status == -ENOMEM) {
+			sedcli_printf(LOG_ERR, "No memory\n");
+		} else {
+			sedcli_printf(LOG_ERR, "Unknown error\n");
+		}
+	 } else {
```
AndrzejJakowski: please fix indentation

src/sedcli_main.c
```diff
@@ -329,10 +329,30 @@ static int get_password(struct sed_key *pwd)
...
+		 sedcli_printf(LOG_INFO, "%s\n", sed_status);
+	 }
+}
```
AndrzejJakowski, line 354: please fix indentation

src/sedcli_main.c
```diff
@@ -329,10 +329,30 @@ static int get_password(struct sed_key *pwd)
 	return 0;
 }
 
+static void sed_status_ret(int status)
```
AndrzejJakowski: please use more descriptive name e.g. print_status_info

src/sedcli_main.c
```diff
@@ -329,10 +329,30 @@ static int get_password(struct sed_key *pwd)
...
+			sedcli_printf(LOG_ERR, "Invalid parameter\n");
+		} else if (status == -ENODEV) {
+			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
+		} else if (status == -ENOMEM) {
+			sedcli_printf(LOG_ERR, "No memory\n");
+		} else {
+			sedcli_printf(LOG_ERR, "Unknown error\n");
+		}
+	 } else {
+		 sedcli_printf(LOG_INFO, "%s\n", sed_status);
```
AndrzejJakowski: please fix indentation

AndrzejJakowski (changes requested):

src/sedcli_main.c
RevanthRajashekar: This else is for the top most if 
also, I use the 'set list' feature on vim to check the indentation, it shows no issue

AndrzejJakowski: > This else is for the top most if
> also, I use the 'set list' feature on vim to check the indentation, it shows no issue

there extra single whitespace before '}', and before 'sedcli_print...' and before '}'

src/sedcli_main.c
RevanthRajashekar: done

AndrzejJakowski: > done

Compilation erros:

sedcli_main.c: In function ‘handle_ownership’:
sedcli_main.c:387:2: error: implicit declaration of function ‘sed_status_ret’ [-Werror=implicit-function-declaration]
  sed_status_ret(ret);
  ^
sedcli_main.c: At top level:
cc1: error: unrecognized command line option "-Wno-address-of-packed-member" [-Werror]
cc1: all warnings being treated as errors

RevanthRajashekar: sorry its the end of the week effect :(
done

src/sedcli_main.c
```diff
@@ -329,10 +329,30 @@ static int get_password(struct sed_key *pwd)
...
+			sedcli_printf(LOG_ERR, "Invalid parameter\n");
+		} else if (status == -ENODEV) {
+			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
+		} else if (status == -ENOMEM) {
+			sedcli_printf(LOG_ERR, "No memory\n");
+		} else {
+			sedcli_printf(LOG_ERR, "Unknown error\n");
+		}
+	} else {
+		sedcli_printf(LOG_INFO, "%s\n", sed_status);
```
AndrzejJakowski: Last minute realization is that only successful status should be printed to stdout, other statuses (i.e errors) should be printed to stderr

RevanthRajashekar: I think other messages like 'Invalid Parameter', 'Not Authorised' or 'SP Busy' etc need not have to be errors, they are just indicating the Driver State, its saying that the SP is busy, or the user is not authorized. The code isn't returning any error right..?

AndrzejJakowski: > I think other messages like 'Invalid Parameter', 'Not Authorised' or 'SP Busy' etc need not have to be errors, they are just indicating the Driver State, its saying that the SP is busy, or the user is not authorized. The code isn't returning any error right..?

TCG_Storage_Architecture_Core_Spec_v2.01_r1.00; chapter 5.1.5 defines status code and its interpretation. I read Not authorized as an error/

RevanthRajashekar: done

src/sedcli_main.c
```diff
@@ -329,10 +331,33 @@ static int get_password(struct sed_key *pwd)
...
+			sedcli_printf(LOG_ERR, "Invalid parameter\n");
+		} else if (status == -ENODEV) {
+			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
+		} else if (status == -ENOMEM) {
+			sedcli_printf(LOG_ERR, "No memory\n");
+		} else {
+			sedcli_printf(LOG_ERR, "Unknown error\n");
+		}
+	} else {
+		if (strncmp(sed_status, SED_SUCCESS, strlen(SED_SUCCESS)) == 0)
```
AndrzejJakowski: Instead of performing strncmp operation here I would expect to have enum containing all opal status codes defined in libsed.h. In this function you would simply perform arithmetic comparison instead of string comparison.

PR #9, opened 2019-10-28
sedcli: Resync SEDCLI Locktype with Opal Driver

RevanthRajashekar:
This patch resyncs sedcli locktype with opal driver.
Also, this patch updates the code wherever required
inaccordance.

### DIFF
--- a/doc/sedcli.8
+++ b/doc/sedcli.8
@@ -58,9 +58,15 @@ Activates Locking SP.
 .IP "\fB\-R -d <device> [-i]\fR or \fB\-\-revert --device <device> [--psid]\fR"
 Reverts TPer to Manufactured-Inactivate state using either SID or PSID authority.
 
-.IP "\fB\-L -d <device> -t <locktype>\fR or \fB\-\-lock-unlock --device <device> --locktype {RW|RO|WO|UNLOCK}\fR"
-Changes the lock state for the device. There is a possibility to setup Read lock,
-Write lock and Read-Write lock and unlock the device.
+.IP "\fB\-L -d <device> -t <accesstype>\fR or \fB\-\-lock-unlock --device <device> --accesstype {RO|RW|LK}\fR"
+.IP
+Changes the lock state for the device. Following access modes are available:
+.IP
+1. Read-Only(RO) - user can only read the data from the disk.
+.IP
+2. Read-Write(RW) - user can read/write data from/to the disk.
+.IP
+3. Locked(LK) - data is locked, user can NOT read/write data from/to the disk.
 
 .IP "\fB\-P -d <device>\fR or \fB\-\-set-password --device <device>\fR"
 Updates password for Admin1 authority in Locking SP.
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -13,11 +13,10 @@
 
 #define SED_MAX_KEY_LEN (256)
 
-enum SED_LOCK_TYPE {
-	SED_NO_LOCK = 0x00,
-	SED_READ_LOCK = 0x01,
-	SED_WRITE_LOCK = 0x02,
-	SED_READ_WRITE_LOCK = (SED_READ_LOCK | SED_WRITE_LOCK),
+enum SED_ACCESS_TYPE {
+	SED_RO_ACCESS = 1 << 0,
+	SED_RW_ACCESS = 1 << 1,
+	SED_NO_ACCESS = 1 << 2,
 };
 
 struct sed_device;
@@ -86,7 +85,7 @@ int sed_setup_global_range(struct sed_device *dev, const struct sed_key *key);
 /**
  *
  */
-int sed_lock_unlock(struct sed_device *dev, const struct sed_key *key, enum SED_LOCK_TYPE lock_type);
+int sed_lock_unlock(struct sed_device *dev, const struct sed_key *key, enum SED_ACCESS_TYPE lock_type);
 
 /**
  *
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -390,18 +390,14 @@ static int opal_rw_lock(struct opal_device *dev, uint8_t *lr_buff, size_t len,
 		return -ERANGE;
 
 	switch (l_state) {
-	case SED_NO_LOCK:
-		*rl = *wl = 0;
-		break;
-	case SED_READ_LOCK:
-		*rl = 1;
-		*wl = 0;
-		break;
-	case SED_WRITE_LOCK:
+	case SED_RO_ACCESS:
 		*rl = 0;
 		*wl = 1;
 		break;
-	case SED_READ_WRITE_LOCK:
+	case SED_RW_ACCESS:
+		*rl = *wl = 0;
+		break;
+	case SED_NO_ACCESS:
 		*rl = *wl = 1;
 		break;
 	default:
@@ -1723,15 +1719,14 @@ int opal_revertlsp_pt(struct sed_device *dev, const struct sed_key *key, bool ke
 }
 
 int opal_add_usr_to_lr_pt(struct sed_device *dev, const char *key, uint8_t key_len,
-		const char *usr, enum SED_LOCK_TYPE lock_type, uint8_t lr)
+		const char *usr, enum SED_ACCESS_TYPE lock_type, uint8_t lr)
 {
 	struct sed_key disk_key;
 	int ret = 0;
 	struct opal_device *opal_dev;
-	uint32_t who;
+	uint32_t who, lk_type = lock_type;
 
-	if (lock_type > SED_READ_WRITE_LOCK || lock_type < SED_NO_LOCK || key == NULL
-		|| key_len == 0 || usr == NULL) {
+	if (lk_type > SED_NO_ACCESS || key == NULL || key_len == 0 || usr == NULL) {
 		SEDCLI_DEBUG_MSG("Need to supply user, lock type and password!\n");
 		return -EINVAL;
 	}
@@ -1750,7 +1745,7 @@ int opal_add_usr_to_lr_pt(struct sed_device *dev, const char *key, uint8_t key_l
 	if (ret)
 		goto end_sessn;
 
-	ret = add_usr_to_lr(dev->fd, opal_dev, lock_type, lr, who);
+	ret = add_usr_to_lr(dev->fd, opal_dev, lk_type, lr, who);
 
 end_sessn:
 	opal_end_session(dev->fd, opal_dev);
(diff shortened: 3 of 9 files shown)


AndrzejJakowski:
Please also update the documentation: help for lock-unlock command and manpage

RevanthRajashekar:
Done :)

RevanthRajashekar:
Man-page has been updated with detailed info. about the lock states

doc/sedcli.8
```diff
@@ -58,9 +58,17 @@ Activates Locking SP.
 .IP "\fB\-R -d <device> [-i]\fR or \fB\-\-revert --device <device> [--psid]\fR"
 Reverts TPer to Manufactured-Inactivate state using either SID or PSID authority.
 
-.IP "\fB\-L -d <device> -t <locktype>\fR or \fB\-\-lock-unlock --device <device> --locktype {RW|RO|WO|UNLOCK}\fR"
+.IP "\fB\-L -d <device> -t <locktype>\fR or \fB\-\-lock-unlock --device <device> --locktype {RO|RW|LK}\fR"
```
AndrzejJakowski: suggest to change the name of the parameter from locktype to accesstype.

RevanthRajashekar: People already familiar with the opal parameters or driver code may find it confusing with the nomenclature if we change it

doc/sedcli.8
```diff
@@ -58,9 +58,17 @@ Activates Locking SP.
...
 Changes the lock state for the device. There is a possibility to setup Read lock,
```
AndrzejJakowski: Changes the lock state for the device. Following access modes are available: 
1. Read-Only (RO) - user can only read data from the disk.
2. Read-Write (RW) - user can read/write data from/to the disk
3. Locked (LK) - data is locked. It is not possible to read or write data from/to the disk,

src/lib/include/libsed.h
```diff
@@ -13,11 +13,16 @@
...
+	SED_RW_SHIFT = 1,
+	SED_LK_SHIFT = 2,
+};
+
 enum SED_LOCK_TYPE {
-	SED_NO_LOCK = 0x00,
-	SED_READ_LOCK = 0x01,
-	SED_WRITE_LOCK = 0x02,
-	SED_READ_WRITE_LOCK = (SED_READ_LOCK | SED_WRITE_LOCK),
+	SED_READ_LOCK = 1 << SED_RO_SHIFT,
```
AndrzejJakowski: This is confusing to me. I read SED_RO_SHIFT as SED Read-only shift and then it is used to define value for SED READ_LOCK. Read-Only meand that read_lock is set to 0 and write lock is set to 1.

src/lib/nvme_pt_ioctl.c
```diff
@@ -1728,10 +1724,9 @@ int opal_add_usr_to_lr_pt(struct sed_device *dev, const char *key, uint8_t key_l
 	struct sed_key disk_key;
 	int ret = 0;
 	struct opal_device *opal_dev;
-	uint32_t who;
+	uint32_t who, lk_type = lock_type;
```
AndrzejJakowski, line 1727: why do we need lk_type variable? can we reuse lock_type variable ?

src/lib/include/libsed.h
RevanthRajashekar: Here I'm trying to define by how many bits does 1 need to be shifted in order to define the lock
eg; SED_RO_SHIIFT = 0;
SED_READ_LOCK = 1 << 0; -> SED_READ_LOCK = 1 and so on

src/lib/nvme_pt_ioctl.c
```diff
@@ -1847,8 +1842,9 @@ int opal_lock_unlock_pt(struct sed_device *dev, const struct sed_key *key,
 {
 	int ret = 0;
 	struct opal_device *opal_dev;
+	uint32_t lk_type = lock_type;
```
AndrzejJakowski, line 1845: why do we need to copy lock_type to  lk_type variable? can we use lock_type instead

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: The reason for assigning it to a uint32_t type is that we needn't have to check for the negative value being passed. Only one check condition is sufficient.

src/lib/sed_ioctl.c
```diff
@@ -58,15 +58,21 @@ static int do_generic_lkul(int fd, const struct sed_key *key,
 					bool sum, unsigned long ioctl_cmd)
 {
 	struct opal_lock_unlock oln = { };
+	uint32_t opal_user = user, opal_locktype = lock_type;
 
-	if (!sum || key == NULL || key->len == 0) {
-		SEDCLI_DEBUG_MSG("Need to supply user, lock type and password!\n");
+	if (key == NULL || key->len == 0) {
```
AndrzejJakowski: !sum check is missing. was it intended?

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: The reason for assigning it to a uint32_t type is that we needn't have to check for the negative value being passed. Only one check condition is sufficient.

src/lib/sed_ioctl.c
```diff
@@ -58,15 +58,21 @@ static int do_generic_lkul(int fd, const struct sed_key *key,
 					bool sum, unsigned long ioctl_cmd)
 {
 	struct opal_lock_unlock oln = { };
+	uint32_t opal_user = user, opal_locktype = lock_type;
```
AndrzejJakowski, line 61: why do we need to copy user and lock_type to local variables?

src/sedcli_main.c
```diff
@@ -63,7 +63,7 @@ static cli_option reverttper_opts[] = {
 
 static cli_option lock_unlock_opts[] = {
 	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
-	{'t', "locktype", "String specifying how to lock/unlock drive. Allowed values: RW/RO/WO/UNLOCK", 1, "FMT", CLI_OPTION_REQUIRED},
+	{'t', "locktype", "String specifying how to lock/unlock drive. Allowed values: RO/RW/LK", 1, "FMT", CLI_OPTION_REQUIRED},
```
AndrzejJakowski: please change parameter name locktype to accesstype. PLease update description

AndrzejJakowski (changes requested):
Please fix review comments

src/lib/include/libsed.h
AndrzejJakowski: I understand bit shift operations. Chosen naming is confusing. Read-Only basically set Write-lock not Read-Lock

doc/sedcli.8
AndrzejJakowski: We are not changing kernel naming. We propose utility and library that doesn't have the same confusion like kernel has.

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: Ok, make sense

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: ok, makes sense

src/lib/include/libsed.h
RevanthRajashekar: Can you please suggest a good naming convention for this...

AndrzejJakowski: Something like SED_RO_ACCESS, SED_RW_ACCESS, SED_NO_ACCESS

doc/sedcli.8
AndrzejJakowski: Please update the manpage to above content

RevanthRajashekar:
Done. Please check once

doc/sedcli.8
RevanthRajashekar: Have been changed :)

src/sedcli_main.c
```diff
@@ -63,7 +63,7 @@ static cli_option reverttper_opts[] = {
 
 static cli_option lock_unlock_opts[] = {
 	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
-	{'t', "locktype", "String specifying how to lock/unlock drive. Allowed values: RW/RO/WO/UNLOCK", 1, "FMT", CLI_OPTION_REQUIRED},
+	{'t', "accesstype", "String specifying access type to the drive. Allowed values: RO/RW/LK", 1, "FMT", CLI_OPTION_REQUIRED},
```
AndrzejJakowski: Please change description to "String specifying access type to the data on drive. Allowed ..."

doc/sedcli.8
AndrzejJakowski: Can you please change description to the suggest description

RevanthRajashekar:
done

RevanthRajashekar:
done

doc/sedcli.8
```diff
@@ -58,9 +58,17 @@ Activates Locking SP.
 .IP "\fB\-R -d <device> [-i]\fR or \fB\-\-revert --device <device> [--psid]\fR"
 Reverts TPer to Manufactured-Inactivate state using either SID or PSID authority.
 
-.IP "\fB\-L -d <device> -t <locktype>\fR or \fB\-\-lock-unlock --device <device> --locktype {RW|RO|WO|UNLOCK}\fR"
+.IP "\fB\-L -d <device> -t <accesstype>\fR or \fB\-\-lock-unlock --device <device> --accesstype {RO|RW|LK}\fR"
 Changes the lock state for the device. There is a possibility to setup Read lock,
```
AndrzejJakowski: Please get rid of following sentences:
"Changes the lock state for the device. There is a possibility to setup Read lock, Write lock and Read-Write lock and unlock the device."

src/sedcli_main.c
```diff
@@ -182,10 +182,9 @@ struct sedcli_options {
 static struct sedcli_options *opts = NULL;
 
 static char *allowed_lock_type[] = {
-	[SED_NO_LOCK] = "UNLOCK",
-	[SED_READ_LOCK] = "WO",
-	[SED_WRITE_LOCK] = "RO",
-	[SED_READ_WRITE_LOCK] = "RW",
+	[0] = "RO",
```
AndrzejJakowski: Suggest to intialize w/o specifying index, to make it less error-prone:
static char *allowed_lock_type[] = { "RO", "RW", "LK" };

RevanthRajashekar:
Done

PR #10, opened 2019-10-28
sedcli: Basic sedcli test enhancement

RevanthRajashekar:
This patch adds the following enhancements to the existing test:
1. Read/RW/Lock lock the Disk
2. Test if these locks function properly by reading/writing from/to the disk
3. Change the admin-1 password
4. Check if the admin-1 password was successfully changed and the old password never work    s
5. Revert TPer with SID & PSID authority

Also, this patch ensures if the user provides device path & device PSID

### DIFF
--- a/test/test_sedcli_basic.py
+++ b/test/test_sedcli_basic.py
@@ -6,6 +6,9 @@
 Tests end-user scenarios for sedcli binary.
 This suite needs to be run against Python 3.x and the sedcli binary must be
 in your $PATH.
+The user must provide:
+	1. Device Path
+	2. Device PSID
 """
 
 import os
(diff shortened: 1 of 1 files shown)


test/test_sedcli_basic.py
```diff
@@ -61,8 +81,76 @@ def test_initial_setup(self):
 		self.assertEqual(pipe.returncode, 0, "sedcli returned error while setting RLE and WLE on global range:\n" + str(errs))
 		self.assertEqual(len(errs), 0, "sedcli returned success code but error message has been printed on stderr:\n" + str(errs))
 
+		# Read Lock Disk
```
AndrzejJakowski, line 81: Could we break this out into multiple test cases, e.g. 
1. test_initial_provision - test provisioning of the drive for security
2. test_lock - tests if locking works correctly that is this is locked and data can't be read or written,
Suggestion is to define smaller granularity tests allowing a person to run them individually instead of having one test testing everything.

test/test_sedcli_basic.py
```diff
@@ -27,6 +30,23 @@ def gen_rand_key(key_len):
 	letters = string.ascii_lowercase
 	return  ''.join(random.choice(letters) for i in range(key_len)).encode('ascii')
 
+def rw_opal_drive(device_path):
+	# Read from the disk
+	pipe = subprocess.Popen(["dd", "if=" + device_path, "of=/dev/null", "bs=4k", "count=100", "iflag=direct"], stdout = subprocess.PIPE, stdin = subprocess.PIPE, stderr = subprocess.PIPE)
+	err = pipe.communicate()
+	if (pipe.returncode == 0):
```
AndrzejJakowski: I would expect to see here some kind of assertion checking for dd exit code with expected exit code, that is: when drive is locked I would expect that read and write should fail, thus dd command will fail. Test code should have appropriate assert calls to check it.

RevanthRajashekar:
Done :)

PR #12, opened 2019-10-29
Initilize sed_device field to prevent sedcli crash on deinit

AndrzejJakowski:
This patch initializes sed_device members to prevent crash resulting
from freeing randomly initialized priv pointer in deintialization path.


RevanthRajashekar:
Looks good for me :)

PR #13, opened 2019-10-30
Detect shared object install directory for distribution

AndrzejJakowski:
This patch fixes a problem where installation script installed libsed
shared object into directory that wasn't in linker search path causing
sedcli not to run. This patch discovers lib installation directory and
configures build script accordingly.


RevanthRajashekar (approved):

RevanthRajashekar:
This patch fixes the problem.
Thx :)

PR #15, opened 2019-11-06
sedcli: Fix the MAX key length based on spec

RevanthRajashekar:
This patch aims at fixing the maximum allowable key length as per
the opal spec.
Reference: Section 5.3.3.16.2 of
https://trustedcomputinggroup.org/wp-content/uploads/TCG_Storage_Architecture_Core_Spec_v2.01_r1.00.pdf

Also, this patch makes suitable changes in sedcli_main file to be
in-par with this change

### DIFF
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -11,7 +11,7 @@
 #include <stdint.h>
 #include <stdbool.h>
 
-#define SED_MAX_KEY_LEN (256)
+#define SED_MAX_KEY_LEN (32)
 
 enum SED_ACCESS_TYPE {
 	SED_RO_ACCESS = 1 << 0,
--- a/src/sedcli_main.c
+++ b/src/sedcli_main.c
@@ -277,51 +277,52 @@ int setpw_handle_opts(char *opt, char **arg)
 
 static int get_password(struct sed_key *pwd)
 {
-	int c, flag = 0;
+	size_t dest = SED_MAX_KEY_LEN + 2;
+	uint8_t temp[dest];
+	int ret, len;
 
 	echo_disable();
 
-	if (fgets((char *) pwd->key, SED_MAX_KEY_LEN, stdin) == NULL) {
-		sedcli_printf(LOG_INFO, "Error getting password\n");
-		return -EINVAL;
-	}
+	memset(temp, 0, dest);
 
+	if (fgets((char *) temp, dest, stdin) == NULL) {
+		sedcli_printf(LOG_INFO, "Error getting password\n");
+		ret = -EINVAL;
+		goto err;
+	}
 	sedcli_printf(LOG_INFO, "\n");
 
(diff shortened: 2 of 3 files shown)


src/sedcli_main.c
```diff
@@ -277,50 +277,31 @@ int setpw_handle_opts(char *opt, char **arg)
...
 		sedcli_printf(LOG_INFO, "Error getting password\n");
 		return -EINVAL;
 	}
-
 	sedcli_printf(LOG_INFO, "\n");
 
-	pwd->len = strnlen((char *) pwd->key, SED_MAX_KEY_LEN);
-
-	/* Handle case when user entered more characters than allowed max size */
-	if (pwd->len == SED_MAX_KEY_LEN - 1 && pwd->key[pwd->len - 1] != '\n') {
```
AndrzejJakowski, line 294: This logic is needed to make sure that the input buffer has been fully read

src/sedcli_main.c
```diff
@@ -277,50 +277,31 @@ int setpw_handle_opts(char *opt, char **arg)
...
-		pwd->len = 0;
+	if (temp[dest - 2] != '\n' && temp[dest - 2] != '\0') {
+		sedcli_printf(LOG_INFO, "Password too long..!!!\n");
 		echo_enable();
 		return -EINVAL;
 	}
 
+	pwd->len = strnlen((char *) temp, SED_MAX_KEY_LEN);
+
+	memcpy(pwd->key, temp, SED_MAX_KEY_LEN);
```
AndrzejJakowski: temp buffer needs to be securely erased after password is copied to the destination buffer

src/lib/include/libsed.h
```diff
@@ -24,6 +24,7 @@ struct sed_device;
 struct sed_key {
 	uint8_t key[SED_MAX_KEY_LEN];
 	uint8_t len;
+	uint8_t __align[7];
```
AndrzejJakowski: not sure why we need to have explicit padding. This looks to me as error prone because when for some reason SED_MAX_KEY_LEN would need to be changed along with size of __align[] array. wouldn't it be easier to allow compiler to auto pad it? we don't use neither plan to use this data structure to interact with HW...

RevanthRajashekar: this is required to maintain the alignment or else might lead to a buffer overflow

src/sedcli_main.c
RevanthRajashekar: Here we read the input from the user and check if the user has given a pin greater than allowed length, using the following condition. Hence, this shouldn't be needed.

src/sedcli_main.c
RevanthRajashekar: Done

src/lib/include/libsed.h
RevanthRajashekar: Aligning to 8-byte boundary

RevanthRajashekar:
Changes have been implemented :)

PR #16, opened 2019-11-07
sedcli: Make a case insensitive and secured comparison in get_lock_type

RevanthRajashekar:

### DIFF
--- a/src/sedcli_main.c
+++ b/src/sedcli_main.c
@@ -234,7 +234,7 @@ static int get_lock_type(const char *lock_type)
 	int i;
 
 	for (i = 0; i < ARRAY_SIZE(allowed_lock_type); i++) {
-		if (0 == strcmp(allowed_lock_type[i], lock_type)) {
+		if (0 == strncasecmp(allowed_lock_type[i], lock_type, 2)) {
 			return (1 << i);
 		}
 	}


src/sedcli_main.c
```diff
@@ -234,7 +234,7 @@ static int get_lock_type(const char *lock_type)
 	int i;
 
 	for (i = 0; i < ARRAY_SIZE(allowed_lock_type); i++) {
-		if (0 == strcmp(allowed_lock_type[i], lock_type)) {
+		if (0 == strncasecmp(allowed_lock_type[i], lock_type, 2)) {
```
AndrzejJakowski, line 237: Couple of questions:
1. How this change improves security? 
2. why to ignore case sensitivity for user input?

RevanthRajashekar: just like strlen and strnlen, we can check only 2 characters here.
case insensitivity is just to make sure that we are not too strict wrt case of the character given by the user

AndrzejJakowski: I don't think we are strict for user. In the documentation we have indicated what are the allowed values source code is simply obeying this rules. With this change we allow undocumented user input.
Could you describe what is the security risk that you have originally indicated?

RevanthRajashekar: Nothing in particular about the security here. Just thought it might be a good change :)

AndrzejJakowski:
Based on the received information abandoning this pull request

PR #17, opened 2019-11-20
sed-cli: CLI to perform level0 Discovery and provide device info. 

RevanthRajashekar:
This patch aims at providing an cli that helps the user to check
if the device provided by the user is SED-OPAL compliant or not.
This works fine only with NVMe passthru mechanism.


AndrzejJakowski:
1. Please print error in the commands (take ownership, activate, ....) when non-sed drive is detected .
2. Please add description for new command in the manpage

PR #18, opened 2019-12-05
Use NVMe PT by default for accessing Opal

AndrzejJakowski:
This patch modifies behaviour of build configuration script to use NVMe
PT interface by default instead of auto detecting available interfaces
(Linux Opal driver or NVMe PT) and using it. By default NVMe PT
interface is compiled in. If one seeks to use Linux Opal interface it is
possible to force build configuration script to try to use it. If
distribution doesn't provide Linux Opal interface and user forces build
configuration script to use it, script will fallback to using NVMe PT
interface.

Change-Id: I9340644ad80907ea333ab72174cdd058b852a144


RevanthRajashekar:
Looks good

PR #19, opened 2019-12-05
sed-cli: CLI to perform level0 Discovery and provide device info.

RevanthRajashekar:
This patch aims at providing an cli that helps the user to
obtain the device info via level0 discovery.
This works fine only with NVMe passthru mechanism.

### DIFF
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -21,6 +21,42 @@ enum SED_ACCESS_TYPE {
 
 struct sed_device;
 
+struct sed_tper_supported_feat {
+	uint8_t sync_supp :1;
+	uint8_t async_supp :1;
+	uint8_t ack_nak_supp :1;
+	uint8_t buff_mgmt_supp :1;
+	uint8_t stream_supp :1;
+	uint8_t reserved1 :1;
+	uint8_t comid_mgmt_supp :1;
+	uint8_t reserved2:1;
+} __attribute__((__packed__));
+
+struct sed_locking_supported_feat {
+	uint8_t locking_supp:1;
+	uint8_t locking_en:1;
+	uint8_t locked:1;
+	uint8_t media_enc:1;
+	uint8_t mbr_en:1;
+	uint8_t mbr_done:1;
+	uint8_t reserved:2;
+} __attribute__((__packed__));
+
+struct sed_opalv200_supported_feat {
+	uint16_t base_comid;
+	uint16_t comid_num;
+	uint8_t reserved1;
+	uint16_t admin_lp_auth_num;
+	uint16_t user_lp_auth_num;
+	uint8_t reserved2[7];
+} __attribute__((__packed__));
+
+struct sed_opal_level0_discovery {
+	struct sed_tper_supported_feat sed_tper;
+	struct sed_locking_supported_feat sed_locking;
+	struct sed_opalv200_supported_feat sed_opalv200;
+};
+
 struct sed_key {
 	uint8_t key[SED_MAX_KEY_LEN];
 	uint8_t len;
@@ -56,6 +92,11 @@ enum sed_status {
  * operation.
  */
 int sed_init(struct sed_device **dev, const char *dev_path);
+
+/**
+ *
+ */
+int sed_level0_discovery(struct sed_opal_level0_discovery *discv);
 
 /**
  *
--- a/src/sedcli_main.c
+++ b/src/sedcli_main.c
@@ -26,6 +26,7 @@
 
 extern sedcli_printf_t sedcli_printf;
 
+static int sed_discv_handle_opts(char *opt, char **arg);
 static int ownership_handle_opts(char *opt, char **arg);
 static int activatelsp_handle_opts(char*opt, char **arg);
 static int reverttper_handle_opts(char *opt, char **arg);
@@ -33,6 +34,7 @@ static int lock_unlock_handle_opts(char *opt, char **arg);
 static int setup_global_range_handle_opts(char *opt, char **arg);
 static int setpw_handle_opts(char *opt, char **arg);
 
+static int handle_sed_discv(void);
 static int handle_ownership(void);
 static int handle_version(void);
 static int handle_help(void);
@@ -44,6 +46,12 @@ static int handle_setpw(void);
 
 static void echo_enable();
 static void echo_disable();
+
+static cli_option sed_discv_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'f', "format", "Output format: {normal|udev}", 1, "FMT", CLI_OPTION_OPTIONAL_ARG},
+	{0}
+};
 
 static cli_option ownership_opts[] = {
 	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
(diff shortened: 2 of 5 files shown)


src/lib/include/libsed.h
```diff
@@ -13,6 +13,8 @@
 
 #define SED_MAX_KEY_LEN (32)
 
+#define ENOTSUPP 524 /* Device is not SED compliant */
```
AndrzejJakowski: please remove as this define is not used anywhere

src/lib/include/libsed.h
```diff
@@ -57,6 +95,11 @@ enum sed_status {
  */
 int sed_init(struct sed_device **dev, const char *dev_path);
 
+/**
+ *
+ */
+int sed_lvl0_discv(struct sed_opal_level0_discovery *discv);
```
AndrzejJakowski: use full name instead of abbreviation for more clarity:
sed_level0_discovery

src/lib/include/libsed.h
```diff
@@ -21,6 +23,42 @@ enum SED_ACCESS_TYPE {
...
+struct sed_opalv200_supported_feat {
+	uint16_t base_comid;
+	uint16_t comid_num;
+	uint8_t reserved1;
+	uint16_t admin_lp_auth_num;
+	uint16_t user_lp_auth_num;
+	uint8_t reserved2[7];
+} __attribute__((__packed__));
+
+struct sed_opal_level0_discovery {
```
AndrzejJakowski, line 54: sed_level0_discovery

src/sedcli_main.c
```diff
@@ -78,6 +86,17 @@ static cli_option setpw_opts[] = {
 };
 
 static cli_command sedcli_commands[] = {
+	{
+		.name = "sed-discv",
```
AndrzejJakowski: discovery

src/sedcli_main.c
```diff
@@ -45,6 +47,12 @@ static int handle_setpw(void);
 static void echo_enable();
 static void echo_disable();
 
+static cli_option sed_discv_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'f', "format", "Output format: normal|udev", 1, "FMT", CLI_OPTION_OPTIONAL_ARG},
```
AndrzejJakowski: Please use standard notation for single option from many choices: {human|udev}

src/sedcli_main.c
```diff
@@ -199,6 +219,32 @@ static void echo_enable()
 	tcsetattr(1, 0, &term);
 }
 
+enum sed_print_flags {
+	SED_NORMAL = 1 << 0,
```
AndrzejJakowski: just a curiosity question. why is bitfield? do you expect 2 format to be used at one time, through( SED_NORMAL|SED_UDEV)?

src/sedcli_main.c
```diff
@@ -349,6 +395,103 @@ static void print_sed_status(int status)
...
+static void sed_discv_print_normal(struct sed_opal_level0_discovery *discv, const char *dev_path)
+{
+	bool locking_enabled;
+	uint16_t comid;
+
+	locking_enabled = discv->sed_locking.locking_en ? true : false;
+	comid = discv->sed_opalv200.base_comid;
+
+	if (!comid)
+		sedcli_printf(LOG_INFO, "%s: is NOT SED-OPAL Compliant.\n", dev_path);
```
AndrzejJakowski: Maybe you also indicate version of opal being supported?

src/sedcli_main.c
```diff
@@ -349,6 +395,103 @@ static void print_sed_status(int status)
...
+	else
+		sedcli_printf(LOG_INFO, "%s: is SED-OPAL Compliant.\n", dev_path);
+
+	if (locking_enabled)
+		sedcli_printf(LOG_INFO, "%s: is provisioned with Locking Enabled\n", dev_path);
+	else
+		sedcli_printf(LOG_INFO, "%s: is NOT provisioned with Locking Unabled\n", dev_path);
```
AndrzejJakowski: Unabled?

src/sedcli_main.c
```diff
@@ -349,6 +395,103 @@ static void print_sed_status(int status)
...
+
+	sedcli_printf(LOG_INFO, "DEV_SED_COMPATIBLE=%s\n", DEV_SED_COMPATIBLE);
+	sedcli_printf(LOG_INFO, "DEV_SED_LOCKED=%s\n", DEV_SED_LOCKED);
+}
+
+static int handle_sed_discv(void)
+{
+	int ret = 0;
+
+	#ifdef CONFIG_OPAL_DRIVER
```
AndrzejJakowski: I would prefer that preprocessor macro to be in the array that defines actual command. In case of usage of OPAL_DRIVER simply there wouldn't be discovery command

src/sedcli_main.c
```diff
@@ -586,9 +729,12 @@ int main(int argc, char *argv[])
 
 	if (opts == NULL) {
 		sedcli_printf(LOG_ERR, "Failed to allocated memory\n");
-		return -1;
+		return -ENOMEM;
 	}
 
+	/* Set the print format to Normal by default */
+	opts->print_fmt = SED_NORMAL;
```
AndrzejJakowski: this better fits in handle sed_discv_handle_opts

AndrzejJakowski (changes requested):

src/lib/include/libsed.h
RevanthRajashekar: this name is to differentiate this structure from the function name in libsed.h

src/sedcli_main.c
RevanthRajashekar: Nope... This will be easy to add new print format types and use them in switch cases to print in corresponding formats

src/sedcli_main.c
RevanthRajashekar: this requires addition of a new structure (opalv100). Right now I'm trying to just use the structures/feature descriptor available.
Also, I have the idea of extending this in a new pull request (Add new structures and print them in discovery command)

src/sedcli_main.c
RevanthRajashekar: I mean Locking Enable flag is 0

RevanthRajashekar:
Changes have been included with new print statements. Please review the latest changes

src/sedcli_main.c
```diff
@@ -78,6 +86,17 @@ static cli_option setpw_opts[] = {
 };
 
 static cli_command sedcli_commands[] = {
+	{
+		.name = "sed-discovery",
```
AndrzejJakowski: To be consistent with other command definition can you please use just "discovery"  w/o any prefixes.

src/sedcli_main.c
AndrzejJakowski: I see if the intent is to be less complex, my recommendation is to use integers (enum) instead of bitfield

src/sedcli_main.c
```diff
@@ -349,6 +402,116 @@ static void print_sed_status(int status)
...
+
+	ret = sed_init(&dev, opts->dev_path);
+	if (ret) {
+		sedcli_printf(LOG_ERR, "%s: Error initializing device\n", opts->dev_path);
+		return -EINVAL;
+	}
+
+	discv = malloc(sizeof(*discv));
+	if (discv == NULL) {
+		sedcli_printf(LOG_ERR, "Memory not allocated.\n");
```
AndrzejJakowski, line 482: memory leak of memory pointed to by dev pointer

src/sedcli_main.c
```diff
@@ -78,6 +86,17 @@ static cli_option setpw_opts[] = {
 };
 
 static cli_command sedcli_commands[] = {
+	{
+		.name = "discovery",
```
RevanthRajashekar, line 90: changed from sed-discovery to discovery

src/sedcli_main.c
```diff
@@ -199,6 +219,39 @@ static void echo_enable()
 	tcsetattr(1, 0, &term);
 }
 
+enum sed_print_flags {
+	SED_NORMAL,
+	SED_UDEV,
+};
```
RevanthRajashekar, line 225: Changed from bit operation to simple enum

src/sedcli_main.c
```diff
@@ -349,6 +402,118 @@ static void print_sed_status(int status)
...
+	if (ret) {
+		sedcli_printf(LOG_ERR, "%s: Error initializing device\n", opts->dev_path);
+		return -EINVAL;
+	}
+
+	discv = malloc(sizeof(*discv));
+	if (discv == NULL) {
+		sedcli_printf(LOG_ERR, "Memory not allocated.\n");
+		ret = -ENOMEM;
+		goto init_deinit;
```
RevanthRajashekar, line 484: memory leak fixed here

PR #21, opened 2020-01-28
sedcli: Prevent Double free in deinit function

RevanthRajashekar:
The current code leads to an error(double free) when there
is an error during initialization (especially with the nvme-pt I/F).

This patch fixes the double free issue. Also, it was successfully
tested using valgrind tool for any memory leaks.

### DIFF
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -316,6 +316,7 @@ void opal_deinit_pt(struct sed_device *dev)
 {
 	if (dev->fd != 0) {
 		close(dev->fd);
+		dev->fd = 0;
 	}
 
 	if (dev->priv != NULL) {
@@ -325,6 +326,7 @@ void opal_deinit_pt(struct sed_device *dev)
 
 	if (discv != NULL) {
 		free(discv);
+		discv = NULL;
 	}
 
 	opal_parser_deinit();
--- a/src/lib/sed.c
+++ b/src/lib/sed.c
@@ -186,10 +186,10 @@ int  sed_level0_discovery(struct sed_opal_level0_discovery *discv)
 void sed_deinit(struct sed_device *dev)
 {
 	if (dev != NULL) {
-		curr_if->deinit_fn(dev);
+		if (dev->fd)
+			curr_if->deinit_fn(dev);
 
 		memset(dev, 0, sizeof(*dev));
-
 		free(dev);
 	}
 }
--- a/src/lib/sed.c
+++ b/src/lib/sed.c
@@ -186,9 +186,7 @@ int  sed_level0_discovery(struct sed_opal_level0_discovery *discv)
 void sed_deinit(struct sed_device *dev)
 {
 	if (dev != NULL) {
-		if (dev->fd)
-			curr_if->deinit_fn(dev);
-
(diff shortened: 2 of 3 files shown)


src/lib/sed.c
```diff
@@ -185,13 +185,11 @@ int  sed_level0_discovery(struct sed_opal_level0_discovery *discv)
 
 void sed_deinit(struct sed_device *dev)
 {
-	if (dev != NULL) {
+	if (dev->fd)
```
AndrzejJakowski: this function will crash if it will be called in following way:
sed_deint(NULL)

src/lib/nvme_pt_ioctl.c
```diff
@@ -328,6 +328,8 @@ void opal_deinit_pt(struct sed_device *dev)
 	}
 
 	opal_parser_deinit();
+
+	dev->fd = 0;
 }
 
 int opal_init_pt(struct sed_device *dev, const char *device_path)
```
AndrzejJakowski: I think opal_deinit_pt(dev) should be called in every error condition before returning error code -- otherwise there will memory leak in case of error

src/lib/nvme_pt_ioctl.c
```diff
@@ -328,6 +328,8 @@ void opal_deinit_pt(struct sed_device *dev)
 	}
```
AndrzejJakowski, line 330: do we need to set discv pointer to NULL after freeing discv?

src/lib/nvme_pt_ioctl.c
```diff
@@ -328,6 +328,8 @@ void opal_deinit_pt(struct sed_device *dev)
...
 
 	opal_parser_deinit();
+
+	dev->fd = 0;
```
AndrzejJakowski: this should go into if block in lines 317-319

AndrzejJakowski (changes requested):

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: This can be done in a different patch

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: I'm not sure if it makes any difference

RevanthRajashekar: Yeah this can avoid a dangling pointer.

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: Sure. I would like to make sure that we won't miss it.

RevanthRajashekar: I will submit another patch for this...

src/lib/sed.c
```diff
@@ -186,10 +186,10 @@ int  sed_level0_discovery(struct sed_opal_level0_discovery *discv)
 void sed_deinit(struct sed_device *dev)
 {
 	if (dev != NULL) {
-		curr_if->deinit_fn(dev);
+		if (dev->fd)
```
AndrzejJakowski: this if check is redundant -- because opal_deinit_pt already ensures that dev->fd is non-zero. please remove it

PR #22, opened 2020-01-29
libsed: change list_lr to return LRs array instead of printing them

mdabrows:
Libsed function used to get list of OPAL locking ranges would be more useful if it returned LR data to the caller instead of print it on standard output.
This pull request consists of 2 patches:
1. Change list_lr to update provided array with LR information.
2. Add new option to sedcli to print locking ranges status.

### DIFF
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -13,6 +13,8 @@
 
 #define SED_MAX_KEY_LEN (32)
 
+#define SED_OPAL_MAX_LRS 9
+
 enum SED_ACCESS_TYPE {
 	SED_RO_ACCESS = 1 << 0,
 	SED_RW_ACCESS = 1 << 1,
@@ -90,6 +92,21 @@ struct sed_opal_level0_discovery {
 struct sed_key {
 	uint8_t key[SED_MAX_KEY_LEN];
 	uint8_t len;
+};
+
+struct sed_opal_lockingrange {
+	size_t start;
+	size_t length;
+	uint8_t lr_id:4;
+	uint8_t read_locked:1;
+	uint8_t write_locked:1;
+	uint8_t rle:1;
+	uint8_t wle:1;
+};
+
+struct sed_opal_lockingranges {
+	struct sed_opal_lockingrange lrs[SED_OPAL_MAX_LRS];
+	uint8_t lr_num;
 };
 
 enum sed_status {
@@ -184,6 +201,17 @@ int sed_revertlsp(struct sed_device *dev, const struct sed_key *key, bool keep_g
  */
 int sed_setpw(struct sed_device *dev, const struct sed_key *old_key,
 		const struct sed_key *new_key);
+
+/**
+ * Get list of locking ranges.
+ * @param dev			the device to operate on
+ * @param key   		the Admin1 password
+ * @param lrs			array of discovered locking ranges
+ *
+ * @return OPAL_SUCCESS on success
+ */
+int sed_list_lr(struct sed_device *dev, const struct sed_key *key,
+                struct sed_opal_lockingranges *lrs);
 
 /**
  *
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -1580,10 +1580,9 @@ static int opal_read_datastr(int fd, struct opal_device *dev, uint8_t *data, uin
 	return ret;
 }
 
-static int list_lr(int fd, struct opal_device *dev)
+static int list_lr(int fd, struct opal_device *dev, struct sed_opal_lockingranges *lrs)
 {
 	int ret;
-	uint64_t num_ranges;
 	uint8_t uid[OPAL_UID_LENGTH];
 
 	ret = opal_generic_get_column(fd, dev, opal_uid[OPAL_LOCKING_INFO_TABLE_UID],
@@ -1591,12 +1590,16 @@ static int list_lr(int fd, struct opal_device *dev)
 	if (ret)
 		return ret;
 
-	num_ranges = dev->payload.tokens[4]->vals.uint + 1;
+	lrs->lr_num = dev->payload.tokens[4]->vals.uint + 1;
 	SEDCLI_DEBUG_PARAM("The number of ranges discovered is: %ld\n",
-			num_ranges);
+			lrs->lr_num);
+	if (lrs->lr_num > SED_OPAL_MAX_LRS) {
+		lrs->lr_num = SED_OPAL_MAX_LRS;
+	}
+
 	opal_put_all_tokens(dev->payload.tokens, &dev->payload.len);
 
-	for (int i = 0; i < num_ranges; i++) {
+	for (int i = 0; i < lrs->lr_num; i++) {
 		ret = build_lr(uid, sizeof(uid), i);
 		if (ret) {
 			SEDCLI_DEBUG_MSG("Error building locking range\n");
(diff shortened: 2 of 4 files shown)


src/sedcli_main.c
```diff
@@ -85,6 +87,11 @@ static cli_option setpw_opts[] = {
 	{0}
 };
 
+static cli_option listlr_opts[] = {
+        {'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
```
AndrzejJakowski: please use tab for identation instead of spacec

src/sedcli_main.c
```diff
@@ -164,6 +171,17 @@ static cli_command sedcli_commands[] = {
 		.flags = 0,
 		.help = NULL
 	},
+	{
```
AndrzejJakowski: I'm concerned that we can quickly run out of available letters. did you consider adding it to discovery command?

src/sedcli_main.c
```diff
@@ -739,6 +766,53 @@ static int handle_setpw(void)
...
+		return ret;
+	}
+
+	ret = sed_list_lr(dev, &opts->pwd, &lrs);
+	if (ret == 0) {
+		for (int i = 0; i < lrs.lr_num; i++) {
+			struct sed_opal_lockingrange *lr = &lrs.lrs[i];
+
+			printf("\nLR: %d begins: %ld for: %ld\n", lr->lr_id,
+			       lr->start, lr->length);
```
AndrzejJakowski: 1. please use sedcli_printf() wrapper instead of printf() directly
2. please use tabs for indentation (in all the occurences)

AndrzejJakowski (changes requested):

src/lib/nvme_pt_ioctl.c
```diff
@@ -2184,29 +2182,24 @@ int opal_ds_add_anybody_get(struct sed_device *dev, const char *key, uint8_t key
...
-		return ret;
-	}
-
-	ret = opal_start_admin1_lsp_session(dev->fd, opal_dev, &disk_key);
+	ret = opal_start_admin1_lsp_session(dev->fd, opal_dev, key);
 	if (ret)
 		goto end_sessn;
 
-	ret = list_lr(dev->fd, opal_dev);
+	ret = list_lr(dev->fd, opal_dev, lrs);
```
RevanthRajashekar, line 2211: I think it would be better to check the lrs pointer for NULL before it is being passed.

src/sedcli_main.c
```diff
@@ -739,6 +766,53 @@ static int handle_setpw(void)
...
+	int ret;
+	struct sed_opal_lockingrange_arr lrs;
+
+	memset(&lrs, 0, sizeof(struct sed_opal_lockingrange_arr));
+
+	sedcli_printf(LOG_INFO, "Enter Admin1 password: ");
+
+	ret = get_password(&opts->pwd);
+
+	if (ret != 0) {
```
RevanthRajashekar: if (ret) should be fine.
I know we have this if(ret != 0) condition through out the code, which can be changed in the future.

src/sedcli_main.c
```diff
@@ -739,6 +766,53 @@ static int handle_setpw(void)
...
+	}
+
+	ret = sed_init(&dev, opts->dev_path);
+	if (ret) {
+		sedcli_printf(LOG_ERR, "Error in initializing the dev: %s\n", opts->dev_path);
+		return ret;
+	}
+
+	ret = sed_list_lr(dev, &opts->pwd, &lrs);
+	if (ret == 0) {
```
RevanthRajashekar: if (!ret)

src/sedcli_main.c
mdabrows: I'm not sure if list of locking ranges would be expected in the level 0 discovery output. Maybe discovery option should be rephrased to "disk details" or something like that?

AndrzejJakowski: It was not  original intention for sedcli --discovery to exactly match opal level 0 discovery (even though current description says so:)). My suggestion would be to leave command name as is, that is sedcli --discovery and consider following:
1. add optional parameter --type with following choices: level0,ranges. syntax for the discovery command would look as follows:
sedcli --discovery --device <DEVICE> [--format {**normal** | udev}] [--type {**level0** | range}]
2. "type" and "format" are both optional; default selections are in bold
3. "type" can accept single option or be defined as bitmap, e.g. user can invoke it specifying multiple types and both would be printed, e.g. --type level0,range will print both discovery information for level0 and locking range
4. update documentation in help and manpage...
let me know if that makes sense

src/lib/nvme_pt_ioctl.c
```diff
@@ -1580,23 +1580,22 @@ static int opal_read_datastr(int fd, struct opal_device *dev, uint8_t *data, uin
 	return ret;
 }
 
-static int list_lr(int fd, struct opal_device *dev)
+static int list_lr(int fd, struct opal_device *dev, struct sed_opal_lockingrange_arr *lrs)
```
AndrzejJakowski: opal spec supports Global range plus at least 8 configurable ranges. This function will crash for a device having 11>= total ranges. This function at least should have protection that data will not be written outside of sed_opal_lockingrange_arr.

src/lib/include/libsed.h
```diff
@@ -13,6 +13,8 @@
 
 #define SED_MAX_KEY_LEN (32)
 
+#define SED_OPAL_MAX_LRS 9
```
AndrzejJakowski, line 16: this is confusing to: here SED_OPAL_MAX_LRS define is set to 9 which suggests to me that it is reflecting minimal required total number of ranges (spec: at least 8 locking ranges + 1 global range).
Below in array definition you add 1. Can you explain why?

src/lib/include/libsed.h
```diff
@@ -92,6 +94,21 @@ struct sed_key {
 	uint8_t len;
 };
 
+struct sed_opal_lockingrange {
+	size_t start;
+	size_t length;
+	uint8_t lr_id;
+	uint8_t read_locked;
```
AndrzejJakowski: read_locked, write_locked, rle, wle AFAIK are single bits. we could space optimize here a little bit ....

src/lib/include/libsed.h
```diff
@@ -185,6 +202,17 @@ int sed_revertlsp(struct sed_device *dev, const struct sed_key *key, bool keep_g
...
+/**
+ * Get list of locking ranges.
+ * @param dev			the device to operate on
+ * @param key   		the Admin1 password
+ * @param lrs			array of discovered locking ranges
+ *
+ * @return OPAL_SUCCESS on success
+ */
+int sed_list_lr(struct sed_device *dev, const struct sed_key *key,
+                struct sed_opal_lockingrange_arr *lrs);
```
AndrzejJakowski: it is good programming practice when passing pointer to first element of array to also pass its size, so function can check if data has not been accessed outside of array boundary

AndrzejJakowski (changes requested):

src/lib/include/libsed.h
mdabrows: I can't find any reason for this so I removed this one additional LR.

src/lib/include/libsed.h
mdabrows: Done

src/lib/include/libsed.h
mdabrows: lrs is a pointer to structure not to first element. Maybe name of this structure is confusing, so I changed it to sed_opal_lockingranges.

AndrzejJakowski: correct -- this is a structure not an array. 
but problem exists in list_lr() function in nvme_pt_ioctl.c
lrs->lr_num can be set to number greater than SED_OPAL_MAX_LRS, e.g. 10
later in that same function you loop:
for (int i = 0; i < lrs->lr_num; i++) {
....
//take reference to item in the array
struct sed_opal_lockingrange *lr = &lrs->lrs[i];
// and update that item
lr->lr_id = i;
}

that loop will for i>=9 will start writing outside of array boundary. this needs to be fixed

mdabrows: I think the issue has been resolved in the newest version by limiting lrs->lr_num to SED_OPAL_MAX_LRS.

AndrzejJakowski (approved):

PR #24, opened 2020-02-05
sedcli: enhancements in get_password

AndrzejJakowski:
This patch introduces following changes:
 * min and max password length requirements can be passed to
get_password() function. It will reject invalid input
 * sedcli from now on accepts password min 8 characters in length
 * get_password() has been moved to separate compilation unit allowing
its reuse in different binaries

### DIFF
--- a/src/sedcli_util.c
+++ b/src/sedcli_util.c
@@ -1,0 +1,79 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <termios.h>
+#include <errno.h>
+
+#include <sys/syslog.h>
+
+#include <libsed.h>
+
+#include "argp.h"
+
+static struct termios term;
+extern sedcli_printf_t sedcli_printf;
+
+static void echo_disable()
+{
+	tcgetattr(1, &term);
+	term.c_cc[VMIN] = 1;
+	term.c_lflag &= ~(ECHO | ICANON);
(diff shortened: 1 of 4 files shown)


src/sedcli_util.c
```diff
@@ -0,0 +1,79 @@
...
+	tcsetattr(1, 0, &term);
+}
+
+static void echo_enable()
+{
+	term.c_lflag |= ECHO | ICANON;
+	tcsetattr(1, 0, &term);
+}
+
+int get_password(char *pwd, uint8_t *len, int min, int max)
```
RevanthRajashekar, line 29: Do any of your future features have min and max value apart from SED_MIN_KEY_LEN and SED_MAX_KEY_LEN respectively..?

RevanthRajashekar: we have the **sed_key** struct defined under lib/include/libsed.h.
Wouldn't it be a better idea to just pass the **sed_key** struct address rather than getting pwd and len individually.

AndrzejJakowski: 1. yep. password protected backup file has slightly different requirements for password length than manual sedcli utility (min=10 max =64). 
2. because of #1 I couldn't reuse sed_key data structure because buffer for key is too small

RevanthRajashekar (approved):
Looks good

PR #25, opened 2020-02-06
libsed: expose get_msid function in libsed library

AndrzejJakowski:
This patch introduces get_msid function to libsed library, so it is
possible to read manufacturer default password.


RevanthRajashekar (approved):
Looks good

PR #26, opened 2020-02-06
libsed: allow changing pin for selected authority

AndrzejJakowski:
So far, it was only possible to change pin for ADMIN1 authority. This
patch extends setpw() function allowing to select authority for which
password will be changed.

### DIFF
--- a/src/lib/nvme_pt_ioctl.h
+++ b/src/lib/nvme_pt_ioctl.h
@@ -375,7 +375,7 @@ int opal_setuplr_pt(struct sed_device *dev, const char *key, uint8_t key_len,
 int opal_lock_unlock_pt(struct sed_device *dev, const struct sed_key *key,
 		enum SED_ACCESS_TYPE lock_type);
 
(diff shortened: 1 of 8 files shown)


src/lib/nvme_pt_ioctl.h
```diff
@@ -375,7 +375,7 @@ int opal_setuplr_pt(struct sed_device *dev, const char *key, uint8_t key_len,
 int opal_lock_unlock_pt(struct sed_device *dev, const struct sed_key *key,
 		enum SED_ACCESS_TYPE lock_type);
 
-int opal_set_pwd_pt(struct sed_device *dev, const struct sed_key *old_key,
+int opal_set_pwd_pt(struct sed_device *dev, enum SED_AUTHORITY auth, const struct sed_key *old_key,
```
RevanthRajashekar, line 378: Need to update the protocol in nvme_pt_ioctl.c as well

AndrzejJakowski: Good catch. posted missing file in the second commit

RevanthRajashekar (approved):
Looks good

PR #33, opened 2020-02-21
sedcli: opal_put_all_tokens must be mandatory to avoid any leak

RevanthRajashekar:
This patch makes sure that opal_put_all_tokens function is called after every
opal_snd_rcv_cmd_parse_chk is done.

### DIFF
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -718,19 +718,19 @@ static int opal_start_anybody_session(int fd, struct opal_device *dev, int sp)
 			opal_uid[OPAL_SM_UID], opal_method[OPAL_STARTSESSION_METHOD_UID]);
 
 	ret = opal_snd_rcv_cmd_parse_chk(fd, dev, false);
-	if (ret == OPAL_SUCCESS) {
-		dev->session.hsn = dev->payload.tokens[4]->vals.uint;
-		dev->session.tsn = dev->payload.tokens[5]->vals.uint;
-	} else {
+	if (ret) {
 		SEDCLI_DEBUG_MSG("Response parsed incorrectly.\n");
-		return ret;
+		goto put_tokens;
 	}
 
+	dev->session.hsn = dev->payload.tokens[4]->vals.uint;
+	dev->session.tsn = dev->payload.tokens[5]->vals.uint;
 	if (dev->session.hsn == 0 && dev->session.tsn == 0) {
 		SEDCLI_DEBUG_MSG("Session couldn't be authenticated\n");
-		return -EINVAL;
+		ret = -EINVAL;
 	}
 
+put_tokens:
 	opal_put_all_tokens(dev->payload.tokens, &dev->payload.len);
 
 	return ret;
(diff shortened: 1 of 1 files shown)


src/lib/nvme_pt_ioctl.c
```diff
@@ -723,14 +723,15 @@ static int opal_start_anybody_session(int fd, struct opal_device *dev, int sp)
 		dev->session.tsn = dev->payload.tokens[5]->vals.uint;
 	} else {
 		SEDCLI_DEBUG_MSG("Response parsed incorrectly.\n");
-		return ret;
+		goto put_tokens;
 	}
 
 	if (dev->session.hsn == 0 && dev->session.tsn == 0) {
```
AndrzejJakowski, line 728: how this condition can be true if we can only get to this if statement when opal_snd_rcv_cmd_parse_chk returns OPAL_SUCCESS?
I think this whole if may not be needed

src/lib/nvme_pt_ioctl.c
```diff
@@ -723,14 +723,15 @@ static int opal_start_anybody_session(int fd, struct opal_device *dev, int sp)
 		dev->session.tsn = dev->payload.tokens[5]->vals.uint;
```
AndrzejJakowski: can we make this a little bit pretty?
ret = opal_snd_rcv_cmd_parse_chk(fd, dev, false);
if (ret != OPAL_SUCCESS) {
    SEDCLI_DEBUG_MSG("Response parsed incorrectly.\n");
   goto put_tokens;
}
dev->session.hsn = dev->payload.tokens[4]->vals.uint;
dev->session.tsn = dev->payload.tokens[5]->vals.uint;



}

src/lib/nvme_pt_ioctl.c
```diff
@@ -949,13 +952,19 @@ static struct opal_req_item opal_generic_get_column_cmd[] = {
...
+
 	opal_generic_get_column_cmd[3].val.uint = start_col;
 	opal_generic_get_column_cmd[7].val.uint = end_col;
 
 	prepare_req_buf(dev, opal_generic_get_column_cmd, ARRAY_SIZE(opal_generic_get_column_cmd),
 			uid, opal_method[OPAL_GET_METHOD_UID]);
 
-	return opal_snd_rcv_cmd_parse_chk(fd, dev, false);
+	ret = opal_snd_rcv_cmd_parse_chk(fd, dev, false);
+	if (ret)
```
AndrzejJakowski: why tokens are not freed when opal_snd_rcv_cmd_parse_chk returns success (status =0)?

src/lib/nvme_pt_ioctl.c
```diff
@@ -970,7 +979,6 @@ static int opal_get_msid(int fd, struct opal_device *dev, uint8_t *key, uint8_t
 		memcpy(key, dev->payload.tokens[4]->pos + jmp, *key_len);
 	} else {
 		SEDCLI_DEBUG_MSG("There is an error in parsing\n");
-		return ret;
```
AndrzejJakowski, line 973: now in case of error opal_put_all_tokens will be called twice:
one in opal_generitc_get_column and second time in opal_get_msid

src/lib/nvme_pt_ioctl.c
```diff
@@ -989,7 +997,7 @@ static int opal_get_lsp_lifecycle(int fd, struct opal_device *dev)
 	lc_status = dev->payload.tokens[4]->vals.uint;
 	if (lc_status != OPAL_MANUFACTURED_INACTIVE) {
 		SEDCLI_DEBUG_MSG("Couldn't determine the status of the Lifecycle state\n");
-		return -ENODEV;
+		ret = -ENODEV;
```
AndrzejJakowski, line 1002: same as above....

AndrzejJakowski (changes requested):

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: the content of the payload is utilized in other functions like, opal_get_msid, opal_get_lsp_lifecycle, get_table_length, list_lr and a few more. If we put all the tokens down here, these functions can loose that payload value.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: yes. will fix it

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: will fix it

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: yep. I noticed that. and here is some concern: before this change the contract was that payload must be freed in the caller. With this change in some cases caller frees the payload in some other cases it is freed in this function. It is really confusing and error prone

RevanthRajashekar: Only in case of success, the caller function will put_all_token.
In case of error, the called func will take care of it

PR #35, opened 2020-02-26
sedcli-kmip: introduction of crypto module

AndrzejJakowski:
This patch introduces crypto module exposing basic cryptographic
function needed for disk key derivation, key wrapping and unwrapping,
and generating random numbers. This patch also adds openssl in version
1.1.1d as git submodule to sedcli project.


RevanthRajashekar:
I think a new branch or a new tag should be created for this and any future works on this...Also, the previous meta data serialize and de-serialize commit should be moved to the new branch/tag

PR #36, opened 2020-02-27
sedcli: Avoid redundancy in start session methods

RevanthRajashekar:
This patch eliminates the redundancy in start session methods
Like the usage of command struct array, preparing session buffer and
validating the session.
With a few minor tweaks in the check for session validity.


AndrzejJakowski:
Nice patch :)

PR #37, opened 2020-02-27
sedcli-kmip: config file parser introduction

AndrzejJakowski:
This patch add support for config file parser and provides example
configuration file.

### DIFF
--- a/src/config_file.c
+++ b/src/config_file.c
@@ -1,0 +1,239 @@
+/*
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <errno.h>
+#include <string.h>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include "config_file.h"
+#include "lib/sedcli_log.h"
+
+#define SEDCLI_CONF_DELIM "="
+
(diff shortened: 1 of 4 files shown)


src/config_file.c
```diff
@@ -0,0 +1,239 @@
...
+
+#include "config_file.h"
+#include "lib/sedcli_log.h"
+
+#define SEDCLI_CONF_DELIM "="
+
+#define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
+
+enum {
+	KMIP_IP = 0,
```
RevanthRajashekar, line 24: By default the first value of an enum is 0.
It shouldn't be necessary to specify this.

src/config_file.c
```diff
@@ -0,0 +1,239 @@
...
+	KMIP_PORT,
+	CLIENT_CERT,
+	CLIENT_KEY,
+	CA_CERT,
+	UNDEFINED
```
RevanthRajashekar, line 29: If there are any plans of extending this enum in future, 
I would suggest introducing a ',' (comma) at the end of the list.

RevanthRajashekar (approved):

src/config_file.c
AndrzejJakowski: Currently no plans to extend this enumeration

src/config_file.c
AndrzejJakowski: I guess it is also ok to explicitly set the value of enum instead of rely on implicit defaults. I will go ahead with this patch

PR #38, opened 2020-02-27
sedcli: opal_put_all_tokens made mandatory to avoid any leak

RevanthRajashekar:
This patch makes sure that opal_put_all_tokens function is called after every
opal_snd_rcv_cmd_parse_chk is done (irrespective of return value).

### DIFF
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -937,15 +937,16 @@ static int opal_get_msid(int fd, struct opal_device *dev, uint8_t *key, uint8_t 
 	int ret = 0;
 
 	ret = opal_generic_get_column(fd, dev, opal_uid[OPAL_C_PIN_MSID_UID], OPAL_PIN, OPAL_PIN);
-	if (ret == OPAL_SUCCESS) {
-		jmp = get_payload_string(dev, 4);
-		*key_len = dev->payload.tokens[4]->len - jmp;
-		memcpy(key, dev->payload.tokens[4]->pos + jmp, *key_len);
-	} else {
-		SEDCLI_DEBUG_MSG("There is an error in parsing\n");
-		return ret;
+	if (ret) {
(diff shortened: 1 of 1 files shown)


src/lib/nvme_pt_ioctl.c
```diff
@@ -958,13 +959,18 @@ static int opal_get_lsp_lifecycle(int fd, struct opal_device *dev)
 
 	ret = opal_generic_get_column(fd, dev, opal_uid[OPAL_LOCKING_SP_UID],
 				      OPAL_LIFECYCLE, OPAL_LIFECYCLE);
+	if (ret) {
+		SEDCLI_DEBUG_MSG("Error parsing payload\n");
+		goto put_tokens;
+	}
 
 	lc_status = dev->payload.tokens[4]->vals.uint;
 	if (lc_status != OPAL_MANUFACTURED_INACTIVE) {
```
AndrzejJakowski, line 968: It appears that Opal spec allows 2 values for lifecycle state column MANUFACTURED_INACTIVE and MANUFACTURED. it seems that MANUFACTURED state will bre treated here as error

AndrzejJakowski (changes requested):

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: That's exactly what we are doing here. If the lc_status is not MANUFACTURED_INACTIVE  then its an error.
(Same as throwing an error when lc_status = MANUFACTURED)
Also , you can refer to page 38 of the Opal application note.
_Results
session[TSN:HSN] <- [ [LifeCycle = Manufactured-Inactive] ]
Table 15 Get – LockingSP – Response_

AndrzejJakowski: I guess I'm questioning why MANUFACTURED state would be treated as an error here if the spec defines is as valid value.

PR #39, opened 2020-02-28
sedcli: Minor tweak in validate_sessn's check condition

RevanthRajashekar:
Changed the operator from && to ||.


AndrzejJakowski:
good catch

PR #42, opened 2020-03-06
sedcli-kmip: adding example certificate file for sedcli-kmip and KMS

AndrzejJakowski:
This patch adds example certificate files that can be used to test
communication with KMS appliance. Client and server certificate are tied
to the same certificate authority (CA). Also this patch adds example
PyKMIP configuration files.


RevanthRajashekar (approved):
Looks Good

PR #45, opened 2020-03-12
sedcli-kmip: added udev rules to auto provision and auto unlock

AndrzejJakowski:
This patch adds udev rules to auro-provision and auto-unlock when NVMe
device is added. This can happen for example when OS boots up or new
NVMe disk is hot-inserted into system.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -134,7 +134,7 @@ install-$(TARGET):
 install-$(TARGET)-kmip: install-$(TARGET)
 	@echo " Installing $(TARGET)-kmip"
 	install -m 755 $(TARGET)-kmip /usr/sbin/$(TARGET)-kmip
-	install -m 644 ../lib/udev/rules.d/63-sedcli.rules /usr/lib/udev/rules.d/
+	install -m 644 ../lib/udev/rules.d/63-sedcli.rules /etc/udev/rules.d/
 	install -m 755 -d /etc/sedcli /etc/sedcli/certs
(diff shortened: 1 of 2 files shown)


src/Makefile
```diff
@@ -134,6 +134,7 @@ install-$(TARGET):
 install-$(TARGET)-kmip: install-$(TARGET)
 	@echo " Installing $(TARGET)-kmip"
 	install -m 755 $(TARGET)-kmip /usr/sbin/$(TARGET)-kmip
+	install -m 644 ../lib/udev/rules.d/63-sedcli.rules /usr/lib/udev/rules.d/
```
RevanthRajashekar: **_/etc/udev/rules.d/_** is reserved for custom made rules.
Also, I don't see **udev** under **/user/lib/** on my linux machine.

RevanthRajashekar: This leads to an error when installing kmip.
 **_Installing sedcli-kmip
install -m 755 sedcli-kmip /usr/sbin/sedcli-kmip
install -m 644 ../lib/udev/rules.d/63-sedcli.rules /usr/lib/udev/rules.d/
install: target '/usr/lib/udev/rules.d/' is not a directory: No such file or directory
Makefile:135: recipe for target 'install-sedcli-kmip' failed
make: *** [install-sedcli-kmip] Error 1_**

AndrzejJakowski: Yep. This makes sense to me. I've updated this to /etc/udev/rules.d/ directory.

RevanthRajashekar (approved):

PR #46, opened 2020-03-13
sedcli: updates in manpage

AndrzejJakowski:
This patch provides description for missing commands (--discovery and
--setup-global-range) in sedcli manpage. It also clarifies description
for --help command explaining possibility of calling it globally or in
context of specific command.


RevanthRajashekar (approved):
Looks good.

PR #49, opened 2020-05-05
Enable/Disable Shadow MBR

RevanthRajashekar:


AndrzejJakowski (commented):
It seems that this functionality is not complete from user perspective. 
I would expect user require to write PBA image into MBR region. With current patch it seems it is not possible. Would it be possbile to modify/extend this patch to also include MBR write procedure?

RevanthRajashekar:
Hey Andrzej,
Will definitely do that. I'll close this pull request and send out new set of patches.

Thanks!

PR #51, opened 2020-05-11
sedcli - Mbr cli enablement

RevanthRajashekar:


src/sedcli_main.c
```diff
@@ -654,6 +700,42 @@ static int handle_setpw(void)
...
+	ret = get_password((char *) opts->pwd.key, &opts->pwd.len,
+			SED_MIN_KEY_LEN, SED_MAX_KEY_LEN);
+	if (ret)
+		return -1;
+
+	ret = sed_init(&dev, opts->dev_path);
+	if (ret) {
+		sedcli_printf(LOG_ERR, "Error in initializing the dev: %s\n",
+			      opts->dev_path);
+                return ret;
```
AndrzejJakowski: ERROR: code indent should use tabs where possible

src/sedcli_main.c
```diff
@@ -81,6 +83,14 @@ static cli_option setpw_opts[] = {
 	{0}
 };
 
+static cli_option mbr_control_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'e', "enable-mbr", "Select MBR Enable column", 0, "FLAG", CLI_OPTION_OPTIONAL_ARG},
```
AndrzejJakowski: this and following option should take parameter indicating desired value by user vs specifying that value through yet another option

AndrzejJakowski: please apply short name "enable" "done"

src/sedcli_main.c
```diff
@@ -91,6 +94,13 @@ static cli_option mbr_control_opts[] = {
 	{0}
 };
 
+static cli_option write_mbr_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'f', "file", "Absolute file path containing data", 1, "FMT", CLI_OPTION_REQUIRED},
```
AndrzejJakowski: why won't it be possible to use relative path to file?

src/sedcli_main.c
```diff
@@ -736,6 +783,55 @@ static int handle_mbr_control(void)
...
+
+	sedcli_printf(LOG_INFO, "Enter Admin1 password: ");
+
+	ret = get_password((char *) opts->pwd.key, &opts->pwd.len,
+			SED_MIN_KEY_LEN, SED_MAX_KEY_LEN);
+	if (ret)
+		return -1;
+
+	mbr_fd = open(opts->file_path, O_RDONLY | O_CLOEXEC);
+	if (fstat(mbr_fd, &mbr_st) == -1) {
```
AndrzejJakowski, line 857: fstat checks could be done before prompting user for password. this would offer better experience

src/sedcli_main.c
```diff
@@ -736,6 +783,55 @@ static int handle_mbr_control(void)
...
+				opts->file_path);
+		ret = -1;
+		goto close_fd;
+	}
+
+	ret = sed_init(&dev, opts->dev_path);
+	if (ret) {
+		sedcli_printf(LOG_ERR, "Error in initializing the dev: %s\n",
+				opts->dev_path);
+		goto close_fd;
```
AndrzejJakowski: in this path file is not munmap

src/sedcli_main.c
```diff
@@ -736,6 +783,55 @@ static int handle_mbr_control(void)
...
+
+	ret = sed_write_shadow_mbr(dev, &opts->pwd, (const uint8_t *)mbr_mmap,
+			mbr_st.st_size, opts->offset);
+
+	print_sed_status(ret);
+
+	sed_deinit(dev);
+
+close_fd:
+	close(mbr_fd);
```
AndrzejJakowski, line 890: missing file munmap

AndrzejJakowski (changes requested):

AndrzejJakowski:
please fix conflict and respond to review comments

RevanthRajashekar:
Changes incorporated...please review it

src/sedcli_main.c
```diff
@@ -293,6 +315,31 @@ int setpw_handle_opts(char *opt, char **arg)
...
+
+bool mbr_enable = false;
+bool mbr_done = false;
+int mbr_control_handle_opts(char *opt, char **arg)
+{
+	if (!strcmp(opt, "device")) {
+		strncpy(opts->dev_path, arg[0], PATH_MAX - 1);
+	} else if (!strcmp(opt, "enable")) {
+		mbr_enable = true;
+		opts->enable = get_mbr_flag(arg[0]);
```
AndrzejJakowski, line 365: getting segmentation fault when providing following input: ./sedcli -M -d /dev/nvme1n1 -e

src/sedcli_main.c
```diff
@@ -654,6 +701,41 @@ static int handle_setpw(void)
...
+	int ret;
+
+	sedcli_printf(LOG_INFO, "Enter Admin1 password: ");
+
+	ret = get_password((char *) opts->pwd.key, &opts->pwd.len,
+			SED_MIN_KEY_LEN, SED_MAX_KEY_LEN);
+	if (ret)
+		return -1;
+
+	ret = sed_init(&dev, opts->dev_path);
```
AndrzejJakowski, line 803: [root@localhost src]# ./sedcli -D -d /dev/nvme1n1 

SED TPER FEATURES SUPPORTED
	Sync Supported        : Y
	Async Supported       : N
	ACK/NAK Supported     : N
	Buffer Mgmt Supported : N
	Streaming Supported   : Y
	ComID Mgmt Supported  : N

SED LOCKING FEATURES SUPPORTED
	Locking Supported : Y
	Locking Enabled   : N
	Locked            : N
	Media Encryption  : Y
	MBR Enabled       : N
	MBR Done          : N

SED GEOMETRY FEATURES SUPPORTED
	Alignment required    : N
	Logical Block Size    : 512
	Alignment Granularity : 1
	Lowest Aligned LBA    : 0

SED DATASTORE FEATURES SUPPORTED
	Max DataStore tables       : 10
	Max size DataStore tables  : 10485760
	DataStore table size align : 4096

SED Opal v1.00 FEATURES SUPPORTED
	Base ComID       : 2048
	Number of ComIDs : 1

SED Opal v2.00 FEATURES SUPPORTED
	Base ComID                      : 2048
	Number of ComIDs                : 1
	Range Crossing Behavior         : 1
	Admin Authorities LSP Supported : 4
	User Authorities LSP Supported  : 9
	Initial PIN                     : 0
	Revert PIN                      : 0

[root@localhost src]# ./sedcli -M -d /dev/nvme1n1 -e TRUE -m TRUE
Enter Admin1 password: 
Password too short..!!
Please provide password min 8 characters long.
[root@localhost src]# test
[root@localhost src]# ./sedcli -M -d /dev/nvme1n1 -e TRUE -m TRUE
Enter Admin1 password: 
Invalid Parameter

src/sedcli_main.c
```diff
@@ -654,6 +701,41 @@ static int handle_setpw(void)
 	return ret;
 }
 
+static int handle_mbr_control(void)
+{
+	struct sed_device *dev = NULL;
+	int ret;
+
+	sedcli_printf(LOG_INFO, "Enter Admin1 password: ");
```
AndrzejJakowski: this should go after sedcli_init to improve user experience.
Now if user specifies invalid disk he is prompted for password and after entering password sedcli errors out because of invalid disk. All necessary checks (including discovery and check if drive is in proper state)  should be done before prompting for password to improve user experience

src/sedcli_main.c
```diff
@@ -90,6 +93,13 @@ static cli_option mbr_control_opts[] = {
 	{0}
 };
 
+static cli_option write_mbr_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'f', "file", "File path containing data", 1, "FMT", CLI_OPTION_REQUIRED},
```
AndrzejJakowski: I suggest following description: "File path containing Pre-boot Application (PBA) image "

AndrzejJakowski:
In addition to those comment please extend manpage to include documentation of added commands

src/sedcli_main.c
RevanthRajashekar: This is because a NULL character is being string compared...will fix it

src/sedcli_main.c
RevanthRajashekar: [root@localhost src]# sedcli -M -d /dev/nvme0n1 -e TRUE -m TRUE
Enter Admin1 password:
Success

Works fine on my system...Can you please re-try again

src/sedcli_main.c
RevanthRajashekar: Will do it

src/sedcli_main.c
RevanthRajashekar: sounds good

src/sedcli_main.c
AndrzejJakowski: I tried on fresh drive from factory -- that's why I included log from discovery
Could you please try the same?

RevanthRajashekar: You need to activate-lsp before you perform any of the MBR control operations.
Once the lsp is activated, the discover command should show 'Locking Enabled : Y'.
Any MBR control operations performed after this should succeed.

AndrzejJakowski: Exactly. Checks should be added in SW making sure that drive is in correct state. And if not error should be returned to user indicating appropriate reason of failure

RevanthRajashekar: I would say that the user should know that the LSP should be activated before the operation is performed. Also, we are providing a Discover Zero command under nvme_pt interface to assist the user with necessary info.
Would it be a good idea to provide a new cli to the user where ownership, activate lsp and setup global locking range are all done in 1 command (initial_setup) ...?

AndrzejJakowski: It is not obvious for user from printed out error (Invalid Parameter) to deduce what needs to be done (take ownership, activate LSP, etc). This needs to be fixed.

Providing aggregate cli command to do multiple things  is a good idea. I thin it is lower priority than MBR control & write because method for intializing device already exists (through multiple commands) while MBR control & write doens't

RevanthRajashekar:
Changes incorporated please review

src/sedcli_main.c
```diff
@@ -293,6 +315,46 @@ int setpw_handle_opts(char *opt, char **arg)
...
+	if (!strncasecmp(mbr_flag, "TRUE", 4))
+		return 1;
+	else if (!strncasecmp(mbr_flag, "FALSE", 5))
+		return 0;
+	else {
+		sedcli_printf(LOG_ERR, "Invalid value given by the user\n");
+		return -EINVAL;
+	}
+
+	return 0;
```
AndrzejJakowski: this return is redundant

src/sedcli_main.c
```diff
@@ -654,6 +716,80 @@ static int handle_setpw(void)
...
+	ret = check_current_levl0_discv();
+	if (ret != -EOPNOTSUPP && ret != 0)
+		goto init_deinit;
+
+	sedcli_printf(LOG_INFO, "Enter Admin1 password: ");
+	ret = get_password((char *) opts->pwd.key, &opts->pwd.len,
+				SED_MIN_KEY_LEN, SED_MAX_KEY_LEN);
+	if (ret) {
+		sedcli_printf(LOG_ERR, "Error getting password\n");
+		return -1;
```
AndrzejJakowski: resource leak in this path

AndrzejJakowski:
Overall looks good. There are couple of minor fixes needed. 
Please also extend manpage by describing added commands to sedcli?

doc/sedcli.8
```diff
@@ -81,6 +81,15 @@ Changes the lock state for the device. Following access modes are available:
 .IP "\fB\-P -d <device>\fR or \fB\-\-set-password --device <device>\fR"
 Updates password for Admin1 authority in Locking SP.
 
+.IP "\fB\-M -d <device> -e <enable> -m <done>\fR or \fB\-\-mbr-control --device <device> --enable {TRUE|FALSE} --done {TRUE|FALSE}\fR"
+.IP
+Enable/Disable MBR Shadow and Set/Unset MBR Done.
+MBR Done is TRUE only when the MBR Shadow is Enabled.
+
+.IP "\fB\-W -d <device> -f <File path containing Pre-boot Application(PBA) image> [-o]\fR or \fB\-\-write-mbr --device <device> --file <File path containing Pre-boot Application(PBA) image> [--offset]\fR"
```
AndrzejJakowski: This would be more concise
"-W -d \<device\> -f \<pba_file\> [-o \<offset\>] or --write-mbr --device \<device\> --file \<pba_file\> [--offset \<offset\> ]"

doc/sedcli.8
```diff
@@ -81,6 +81,15 @@ Changes the lock state for the device. Following access modes are available:
...
+Write data into MBR shadow region(MBR table).
+Enable MBR shadow, for the PBA to take effect.
```
AndrzejJakowski: I don't think it actually enables MBR shadow. it just writes the data into MBR table. enabling is done via --mbr-control command

doc/sedcli.8
RevanthRajashekar: I agree on that

doc/sedcli.8
RevanthRajashekar: What i was trying to convey here is that, the user should have enabled the MBR shadow in-order to let the OS use the data stored in MBR table... maybe I can rephrase the sentence.

doc/sedcli.8
```diff
@@ -81,6 +81,15 @@ Changes the lock state for the device. Following access modes are available:
 .IP "\fB\-P -d <device>\fR or \fB\-\-set-password --device <device>\fR"
 Updates password for Admin1 authority in Locking SP.
 
+.IP "\fB\-M -d <device> -e <enable> -m <done>\fR or \fB\-\-mbr-control --device <device> --enable {TRUE|FALSE} --done {TRUE|FALSE}\fR"
```
AndrzejJakowski: Please indicate which parameters are optional and selection list vs placeholder for arbitrat value. Here is my recommendation:
-M -d \<device\> [-e {TRUE|FALSE}]  [-m {TRUE|FALSE}] or --mbr-control --device \<device\>
       [--enable {TRUE|FALSE}] [--done {TRUE|FALSE}]

doc/sedcli.8
```diff
@@ -81,6 +81,15 @@ Changes the lock state for the device. Following access modes are available:
...
+.IP
+Enable/Disable MBR Shadow and Set/Unset MBR Done.
```
AndrzejJakowski: I think it should be:
"Enable/Disable MBR Shadow and/or Set/Unset MBR Done." because user can do both or either one, correct?

"MBR Done is TRUE only when the MBR Shadow is Enabled." -> "MBR Done update to TRUE will only take effect when MBR Shadow is enabled."

doc/sedcli.8
AndrzejJakowski: yep I see now.

AndrzejJakowski:
Just few small fixes in manpage and I think we are good to go

RevanthRajashekar:
All changes incorporated wrt the manpage update. Please review.

PR #52, opened 2020-05-16
argp: Replace deprecated ftime() calls

knthm:
POSIX.1-2008 obsoletes ftime().

Replace ftime() calls with clock_gettime()
Replace timeb with timespec struct.

Signed-off-by: Kenneth J. Miller <ken@miller.ec>

### DIFF
--- a/src/argp.c
+++ b/src/argp.c
@@ -10,7 +10,6 @@
 #include <stdarg.h>
 #include <stdint.h>
 #include <time.h>
-#include <sys/timeb.h>
 #include <sys/stat.h>
 #include <sys/syslog.h>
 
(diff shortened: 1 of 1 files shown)


RevanthRajashekar (approved):
Looks good for me

src/argp.c
```diff
@@ -592,8 +591,8 @@ int run_command(cli_command *commands, int cmd, int argc, char **argv)
 
 	/* execute command */
 	result = commands[cmd].handle();
-	ftime(&t1);
-	timespan = (1000 * (t1.time - t0.time) + t1.millitm - t0.millitm);
+	clock_gettime(CLOCK_REALTIME, &t1);
+	timespan = (1000 * (t1.tv_sec - t0.tv_sec) + t1.tv_nsec - t0.tv_nsec);
```
AndrzejJakowski: timespan calculation uses incorrect multiplier (1000) for clock_gettime() method -- ftime() returned miliseconds while clock_gettime() returns nanoseconds. Please fix it

AndrzejJakowski (changes requested):
Please fix problem with incorrect multipler

src/argp.c
knthm: You're right of course, fixed in latest push.

Insert some quote about coding tired and reviewing awake. ;)

PR #53, opened 2020-05-29
libsed: nvme_pt: Consider mbr done when enabling/disabling mbr shadow

RevanthRajashekar:
Current behavior:
mbr-control enbale=TRUE and done=TRUE => Discover0 MBR Enabled=Y and MBR
Done=Y.
mbr-control enbale=FLASE => Discover0 MBR Enabled=N and MBRDone=Y.
This is a violation of spec.

Actual behavior:
mbr-control enbale=TRUE and done=TRUE => Discover0 MBR Enabled=Y and MBR
Done = Y.
mbr-control enbale=FLASE => Discover0 MBR Enabled=N and MBRDone=N.

### DIFF
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -2086,6 +2086,10 @@ int opal_shadow_mbr_pt(struct sed_device *dev, const struct sed_key *key,
 	if (ret)
 		goto end_sessn;
 
+	ret = opal_set_mbr_done(dev->fd, opal_dev, enable_disable);
+	if (ret)
+		goto end_sessn;
+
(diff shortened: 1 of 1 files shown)


AndrzejJakowski:
Thx! Couple of comments:
1. There is not spec violation in my understanding - please rephrase patch description
2. Current implementation accepts --enable FALSE and --done TRUE user input but actually ignores --done setting. It should let user know that the input "done" input doesn't take any effect when --enable is TRUE and eityher continue or error out
3. I don't think we should block this input in libarary for following reasons:
 * input is valid -- spec just doesn't care about values
 * better place for this check is in cli

RevanthRajashekar:
The code would be messy considering all the corner cases here.
So I would better leave the way it is.

PR #54, opened 2020-05-29
libsed: Make opal level0 discovery struct sed device context

RevanthRajashekar:
Make level0 discovery struct sed device context rather than an
individual struct avoiding any over-written conditions.

Move the sed_device struct from sed_util library to libsed library.

### DIFF
--- a/src/lib/include/libsed.h
+++ b/src/lib/include/libsed.h
@@ -159,7 +159,8 @@ int sed_init(struct sed_device **dev, const char *dev_path);
 /**
  *
  */
-int sed_level0_discovery(struct sed_opal_level0_discovery *discv);
+int sed_level0_discovery(struct sed_device *dev,
+			struct sed_opal_level0_discovery *discv);
 
 /**
  *
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -198,42 +198,42 @@ static int get_opal_auth_uid(enum SED_AUTHORITY auth)
 	return sed2opal_map[auth];
 }
 
-static void check_tper_feat(struct sed_opal_level0_discovery *discv,
+static void cpy_tper_feat(struct sed_opal_level0_discovery *discv,
 				void *feat)
 {
 	memcpy(&discv->sed_tper, (struct tper_supported_feat *)feat,
 		sizeof(struct tper_supported_feat));
 }
 
-static void check_locking_feat(struct sed_opal_level0_discovery *discv,
+static void cpy_locking_feat(struct sed_opal_level0_discovery *discv,
 				void *feat)
 {
 	memcpy(&discv->sed_locking, (struct locking_supported_feat *)feat,
 		sizeof(struct locking_supported_feat));
 }
 
-static void check_geometry_feat(struct sed_opal_level0_discovery *discv,
+static void cpy_geometry_feat(struct sed_opal_level0_discovery *discv,
 				void *feat)
 {
 	memcpy(&discv->sed_geo, (struct geometry_supported_feat *)feat,
 		sizeof(struct geometry_supported_feat));
 }
 
-static void check_datastr_feat(struct sed_opal_level0_discovery *discv,
+static void cpy_datastr_feat(struct sed_opal_level0_discovery *discv,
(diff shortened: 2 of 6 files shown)


src/lib/include/libsed.h
```diff
@@ -105,6 +103,12 @@ struct sed_opal_level0_discovery {
 	struct sed_opalv200_supported_feat sed_opalv200;
 };
 
+struct sed_device {
```
AndrzejJakowski: I would like to keep that structure private, w/o defining its fields to callers.

src/lib/include/libsed.h
```diff
@@ -159,7 +163,7 @@ int sed_init(struct sed_device **dev, const char *dev_path);
 /**
  *
  */
-int sed_level0_discovery(struct sed_opal_level0_discovery *discv);
```
AndrzejJakowski, line 162: I like original prototype of this function which takes discovery data structure and populates it w/ data.

src/lib/include/libsed.h
```diff
@@ -159,7 +163,7 @@ int sed_init(struct sed_device **dev, const char *dev_path);
...
+int sed_level0_discovery(void);
```
AndrzejJakowski: this seems to have no effect

src/lib/include/libsed.h
RevanthRajashekar: Done

src/lib/include/libsed.h
RevanthRajashekar: Done

src/lib/nvme_pt_ioctl.c
```diff
@@ -414,9 +413,18 @@ int opal_init_pt(struct sed_device *dev, const char *device_path)
...
+				struct sed_opal_level0_discovery *discv)
 {
-	memcpy(discvry, (struct sed_opal_level0_discovery *)discv, sizeof(*discvry));
+	int ret;
+
+	ret = opal_level0_disc_pt(dev);
+	if (ret)
+		return ret;
+
+	memcpy(discv, &dev->discv, sizeof(*discv));
```
AndrzejJakowski, line 428: before doing memcpy you could check if if input is not null

src/lib/nvme_pt_ioctl.c
```diff
@@ -200,42 +198,57 @@ static int get_opal_auth_uid(enum SED_AUTHORITY auth)
 	return sed2opal_map[auth];
 }
 
-static void check_tper_feat(void *feat)
+static void check_tper_feat(struct sed_opal_level0_discovery *discv,
```
AndrzejJakowski: function naming is a little bit confusing: it doesn't check anything it just perform memcpy

src/sedcli_main.c
```diff
@@ -761,37 +751,36 @@ static int handle_setpw(void)
...
+	ret = sed_level0_discovery(dev, &discv);
 	if (ret) {
-		if (ret == -EOPNOTSUPP)
-			sedcli_printf(LOG_ERR, "Command NOT supported for this "
-					"interface.\n");
-		else
-			sedcli_printf(LOG_ERR, "Error in level0 discovery\n");
-		goto free_mem;
+		if (ret == -EOPNOTSUPP) {
+			sedcli_printf(LOG_ERR, "Level0 discovery not supported "
```
AndrzejJakowski: sounds like this is should be logged as LOG_WARNING vs LOG_ERR, because you allow operations

AndrzejJakowski (commented):
Besides this few minor comments this looks good to me

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: yup, will put it in another commit

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: sorry for missing out this check condition

src/sedcli_main.c
RevanthRajashekar: will do it

RevanthRajashekar:
Changes incorporated based on comments. Please review

PR #55, opened 2020-06-05
Mbr control impr

AndrzejJakowski:


src/sedcli_main.c
```diff
@@ -800,6 +800,13 @@ static int handle_mbr_control(void)
 	struct sed_device *dev = NULL;
 	int ret;
 
+	if (!opts->enable && opts->done) {
+		sedcli_printf(LOG_ERR, "Error: disabling MBR shadow and setting "
+				"MBR done doesn't take any effect\n");
+
+		return -EINVAL;
+	}
+
```
RevanthRajashekar, line 809: This error condition check doesn't catch the following case: 
step1: sedcli -M -d /dev/nvme0n1 -e true -m true => discovery MBR_ENABLE = Y and MBR_DONE = Y
step2: sedcli -M -d /dev/nvme0n1 -e false => discovery MBR_ENABLE = F and MBR_DONE = Y

AndrzejJakowski: which case is bad? from what I have heard so far is that case when user simultaneously sets enable=false and done=true is bad.

RevanthRajashekar: In that case, this looks good to me :)

AndrzejJakowski: ok merging then :)

PR #57, opened 2020-06-25
Tper host large pkt sleep

RevanthRajashekar:


src/lib/include/libsed.h
```diff
@@ -96,13 +96,32 @@ struct sed_opalv200_supported_feat {
...
+} __attribute__((__packed__));
+
 struct sed_opal_level0_discovery {
 	struct sed_tper_supported_feat sed_tper;
 	struct sed_locking_supported_feat sed_locking;
 	struct sed_geometry_supported_feat sed_geo;
 	struct sed_datastr_table_supported_feat sed_datastr;
 	struct sed_opalv100_supported_feat sed_opalv100;
 	struct sed_opalv200_supported_feat sed_opalv200;
+	struct sed_tper_properties sed_tper_props;
```
AndrzejJakowski, line 120: Tper properties doesn't belong here.
According to spec it is Level 1 discovery (see page 67 of TCG Storage Architecture Core Spec)

RevanthRajashekar: will remove it and add a new structure

RevanthRajashekar: Done

src/lib/nvme_pt_ioctl.c
```diff
@@ -413,8 +413,10 @@ int opal_init_pt(struct sed_device *dev, const char *device_path)
 	return ret;
 }
 
-int opal_level0_discv_info_pt(struct sed_device *dev,
-				struct sed_opal_level0_discovery *discv)
+static int opal_tper_host_prep(struct sed_device *dev);
```
AndrzejJakowski: prefer this declaration to move to the top of the file.

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,94 @@ static int validate_sessn(struct opal_device *dev)
 	return 0;
 }
 
+static int  parse_tper_host_prop(struct sed_device *device)
```
AndrzejJakowski: please remove extra space between return type and function name

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,94 @@ static int validate_sessn(struct opal_device *dev)
...
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTLIST } },
+
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTNAME } },
+	{ .type = OPAL_BYTES, .len = 16, .val = { .bytes = (uint8_t *)"MaxComPacketSize" } },
+	{ .type = OPAL_U64, .len = 2, .val = { .uint = 2048 } },
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_ENDNAME } },
+
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTNAME } },
+	{ .type = OPAL_BYTES, .len = 24, .val = { .bytes = (uint8_t *)"MaxResponseComPacketSize" } },
+	{ .type = OPAL_U64, .len = 2, .val = { .uint = 4096 } },
```
AndrzejJakowski: is this value correct?

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,94 @@ static int validate_sessn(struct opal_device *dev)
...
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_ENDNAME } },
+
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTNAME } },
+	{ .type = OPAL_BYTES, .len = 13, .val = { .bytes = (uint8_t *)"MaxPacketSize" } },
+	{ .type = OPAL_U64, .len = 2, .val = { .uint = 2028 } },
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_ENDNAME } },
+
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTNAME } },
+	{ .type = OPAL_BYTES, .len = 15, .val = { .bytes = (uint8_t *)"MaxIndTokenSize" } },
+	{ .type = OPAL_U64, .len = 2, .val = { .uint = 1992 } },
```
AndrzejJakowski: how did you derive this value?

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,94 @@ static int validate_sessn(struct opal_device *dev)
 	return 0;
 }
 
+static int  parse_tper_host_prop(struct sed_device *device)
+{
```
AndrzejJakowski, line 742: please fix coding style issues: multiple missing spaces

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: will do it

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,94 @@ static int validate_sessn(struct opal_device *dev)
...
+	struct sed_tper_properties *tper = &device->discv.sed_tper_props;
+	uint64_t val_ptr = sizeof(struct sed_tper_properties);
+	uint64_t val, data_ptr = 0;
+
+	for (int i = 0; i < payload_len; i++) {
+		if(resp_token_match(dev->payload.tokens[i], OPAL_STARTNAME)) {
+			val = dev->payload.tokens[i+2]->vals.uint;
+
+			if (data_ptr < val_ptr)  {
+				memcpy((uint8_t *)tper + data_ptr, &val, sizeof(uint64_t));
```
AndrzejJakowski: Properties are returned as key-value pairs. there needs to be code that matches key (property name) to a specific member in properties data structure.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: will do it

doc/sedcli.8
```diff
@@ -51,7 +51,7 @@ update Admin1 password for Locking SP.
 .SH OPTIONS
 
 .IP "\fB\-D -d <device> [-f {normal|udev}]\fR or \fB\-\-discovery --device <device> [--format {normal|udev}]\fR"
-Performs Level 0 Discovery and prints out info in human-readable (default) or
+Provides level0 discovery information and tper properties of the device. Prints the info in human-readable (default) or
```
AndrzejJakowski: "Performs discovery of the device properties (e.g. through Level 0, Level 1 Opal Discovery)"

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: 1st : the max response com packet size should be changed to 2048 here.
2nd : These values are derived from the Section 3.2.1.2 of Opal SSC Application note.
the spec takes 4k buffer as an example. However, we consider 2k as our buffer size

src/sedcli_main.c
```diff
@@ -104,8 +104,8 @@ static cli_command sedcli_commands[] = {
 	{
 		.name = "discovery",
 		.short_name = 'D',
-		.desc = "Performs SED level0 discovery(works only with NVMe passthru mechanism)",
-		.long_desc = "Performs SED level0 discovery and provides basic info about the device(works only with NVMe passthru mechanism)",
+		.desc = "Performs SED Opal Device discovery(works only with NVMe passthru mechanism)",
+		.long_desc = "Performs SED Opal Device discovery. This command provides level0 discovery info. and tper properties of the device(works only with NVMe passthru mechanism)",
```
AndrzejJakowski: It would be more extensible if you just say: "Performs discovery of device properties (e.g. through Level 0 and Level 1 Discovery)"

AndrzejJakowski (commented):
please update commit desacription to "sedcli" name instead of "sed-cli"

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: Thx. I think these values should match whatever sedcli is currently capable of, for example if sedcli can send packets at max 2048B in size we should advertise this to Tper.

RevanthRajashekar:
Changes incorporated based on the comments and ready for review.

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,128 @@ static int validate_sessn(struct opal_device *dev)
...
+	return 0;
+}
+
+static struct opal_req_item opal_tper_host_prep_cmd[] = {
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTNAME } },
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_HOSTPROPERTIES } },
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTLIST } },
+
+	{ .type = OPAL_U8, .len = 1, .val = { .byte = OPAL_STARTNAME } },
+	{ .type = OPAL_BYTES, .len = 16, .val = { .bytes = (uint8_t *)"MaxComPacketSize" } },
```
AndrzejJakowski: After reading spec wording I think we should not communicate minimum HostProperties to TPer, because it will indicate to Tper that host can't do better than the minimums.
Our intention is to discover whatever TPer is capable of and update and update buffer size etc. to match what Tper advertises to us.

Here is the relevant spec wording:
The TPer uses these host properties when it is constructing responses to be transmitted to the host.
The host MAY omit properties as necessary, depending on the host’s communications capabilities. If
the host omits a property, the value of that property SHALL NOT change from its current value. If the
host specifies a value for a property that does not meet the minimum requirement as defined in Table
168, then the TPer SHALL use the minimum value defined in Table 168 in place of the value supplied
by the host

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,128 @@ static int validate_sessn(struct opal_device *dev)
...
+	struct opal_device *dev = device->priv;
+	int payload_len = dev->payload.len;
+	struct sed_tper_properties *tper = &device->discv.sed_tper_props;
+
+	/* TPer properties are returned as key-value pairs */
+	for (int i = 0; i < payload_len; i++) {
+		if(resp_token_match(dev->payload.tokens[i], OPAL_STARTNAME)) {
+			int jmp = get_payload_string(dev, i + 1);
+			int key_len = dev->payload.tokens[i + 1]->len - jmp;
+			char *key = malloc(key_len);
```
AndrzejJakowski: Current code does following for each name-value pair:
1. calculate key len
2. allocate buffer,
3. process property
4. free memory.

Instead of calculating key_len each time you could assume some max e.g. 64B and preallocate buffer, and reause buffer for each iteration(BTW I think that spec defines max len as 32B ). When you recognize discover property name larger than this buffer code could simply crash indicating that buffer size needs to be larger . This approach could help remove significant complexity from the loop of allocating and deallocating memory and handling errors related to it.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: Yeah. The maxcompacketsize can be removed off from here.

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +733,128 @@ static int validate_sessn(struct opal_device *dev)
...
+			if (!key)
+				return -ENOMEM;
+
+			memset(key, 0, key_len);
+			memcpy(key, dev->payload.tokens[i + 1]->pos + jmp, key_len);
+
+			if (strncmp("MaxComPacketSize", key, key_len) == 0)
```
AndrzejJakowski: this could also be simplfied. One idea if sed_tper_properties is defined as follows:
struct sed_tper_properties {
     struct  {
         char *name;
         uint64_t value;
     } property[23];
};

then this routine could populate property array instead of having massive if else block

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: we should remove HostProperties Parameter from this call not just single name-value pair

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: I'm not just trying to populate here...I'm trying to do a key match here. hence comparing them.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: I agree with that...getting rid of the Host Tper properties is a good idea.
From my interpretation of the spec lines you mentioned, if the HOST properties mentioned doesn't meet the minimum values as mentioned in the table 168, only then will the host properties value get substituted.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: The number of TPer properties returned may vary as spec says. But for time being we can assign a higher value and use it.

RevanthRajashekar:
Changes incorporated based on the feedback. Please review.

src/lib/include/libsed.h
```diff
@@ -96,6 +96,13 @@ struct sed_opalv200_supported_feat {
 	uint8_t reserved2[5];
 } __attribute__((__packed__));
 
+struct sed_tper_properties {
+	struct {
+		char key_name[32];
```
AndrzejJakowski: It would be great if you have defines for 32 and 23 magic numbers. Later you also use 23 in assert call. You should use define there.

src/lib/nvme_pt_ioctl.c
```diff
@@ -727,6 +738,51 @@ static int validate_sessn(struct opal_device *dev)
...
+			memcpy(tper->property[j].key_name,
+				dev->payload.tokens[i + 1]->pos + jmp,
+				key_len);
+			tper->property[j].value =
+					dev->payload.tokens[i + 2]->vals.uint;
+			j++;
+		}
+	}
+
+	return 0;
```
AndrzejJakowski: This function always return 0. why not to have it return void?

src/sedcli_main.c
```diff
@@ -424,64 +424,76 @@ static void print_sed_status(int status)
...
+	sedcli_printf(LOG_INFO, "\tNumber of ComIDs                : %d\n", be16toh(discv->sed_lvl0_discv.sed_opalv200.comid_num));
+	sedcli_printf(LOG_INFO, "\tRange Crossing Behavior         : %d\n", discv->sed_lvl0_discv.sed_opalv200.rangecross_rsvd.range_crossing ? 0 : 1);
+	sedcli_printf(LOG_INFO, "\tAdmin Authorities LSP Supported : %d\n", be16toh(discv->sed_lvl0_discv.sed_opalv200.admin_lp_auth_num));
+	sedcli_printf(LOG_INFO, "\tUser Authorities LSP Supported  : %d\n", be16toh(discv->sed_lvl0_discv.sed_opalv200.user_lp_auth_num));
+	sedcli_printf(LOG_INFO, "\tInitial PIN                     : %d\n", discv->sed_lvl0_discv.sed_opalv200.init_pin);
+	sedcli_printf(LOG_INFO, "\tRevert PIN                      : %d\n", discv->sed_lvl0_discv.sed_opalv200.revert_pin);
+
+	/* Printing TPer Properties */
+	sedcli_printf(LOG_INFO, "\nTPER PROPERTIES\n");
+	for (int i = 0; i < 23; i++) {
```
AndrzejJakowski: instead of "23" either define or array_size should be used

AndrzejJakowski (changes requested):

AndrzejJakowski (commented):
Please add commit description and fix title:
"sed-cli" -> "sedcli"

src/lib/include/libsed.h
RevanthRajashekar: Makes sense. will do it

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: yes!, can be changed to void

src/sedcli_main.c
RevanthRajashekar: will define the value in libsed and use it here

RevanthRajashekar:
Done! Changes incorporated.

PR #59, opened 2020-07-20
Io buffer resize based on tper max compkt sz

RevanthRajashekar:


src/lib/nvme_pt_ioctl.c
```diff
@@ -36,13 +36,15 @@
...
+#define MIN_IO_BUFFER_LEN 2048
 
 struct opal_device {
 	uint16_t comid;
 
-	uint8_t req_buf[IO_BUFFER_LEN];
-	uint8_t resp_buf[IO_BUFFER_LEN];
+	uint8_t *req_buf;
+	uint8_t *resp_buf;
+	uint64_t req_buf_size;
```
AndrzejJakowski, line 49: do you expect buffer size to be 64bit in length?

src/lib/nvme_pt_ioctl.c
```diff
@@ -355,19 +358,70 @@ static int opal_level0_disc_pt(struct sed_device *device)
...
+			return tper->property[j].value;
+	}
+
+	SEDCLI_DEBUG_MSG("Invalid TPer property name!\n");
+
+	return -EINVAL;
+}
+
+static int resize_io_buf(struct opal_device *dev, uint64_t size)
+{
```
AndrzejJakowski, line 400: this function uses whitespace indentation vs /t

src/lib/nvme_pt_ioctl.c
```diff
@@ -355,19 +358,70 @@ static int opal_level0_disc_pt(struct sed_device *device)
...
+               free(dev->req_buf);
+       if (dev->resp_buf != NULL)
+               free(dev->resp_buf);
+
+       dev->req_buf = malloc(sizeof(uint8_t) * size);
+       if (dev->req_buf == NULL)
+               return -ENOMEM;
+       dev->resp_buf = malloc(sizeof(uint8_t) * size);
+       if (dev->resp_buf == NULL)
+               return -ENOMEM;
```
AndrzejJakowski: when code follows this path then req_buf will store pointer to allocated memory., while resp_buf will not. I think this is error prone, because it leaves deallocation to the caller. 
To make things simplier I would suggest to do only one allocation of 2*size. this will simplify function: for example:

....
/* allocate buffer for req and resp*/
ptr = malloc(sizeof(uint8_t) * size * 2);
if (!ptr) {
....
}
memset (ptr, 0, size);
dev->req_buf = &ptr[0];
dev->resp_buf = &ptr[size];

src/lib/nvme_pt_ioctl.c
```diff
@@ -400,14 +454,29 @@ int opal_init_pt(struct sed_device *dev, const char *device_path)
...
+	ret = resize_io_buf(opal_dev, MIN_IO_BUFFER_LEN);
+	if (ret)
+		goto init_deinit;
+
+	ret = opal_dev_discv(dev);
 	if (ret)
 		goto init_deinit;
 
-	SEDCLI_DEBUG_PARAM("The device comid is: %u\n", opal_dev->comid);
+	uint64_t max_com_pkt_sz = tper_prop_to_val(dev, "MaxComPacketSize");
```
AndrzejJakowski: would appreciate define for "MaxComPacketSize" literal

src/lib/opal_parser.c
```diff
@@ -113,6 +113,26 @@ static int append_medium_atom_bytes_header(uint8_t *buf, size_t len, int data_le
 	return 0;
 }
 
+static int append_long_atom_bytes_header(uint8_t *buf, size_t len, int data_len)
+{
```
AndrzejJakowski, line 122: please fix coding style to proper indentation

src/lib/opal_parser.c
```diff
@@ -161,25 +181,28 @@ int append_bytes(uint8_t *buf, size_t len, const uint8_t *src, int src_len)
 
 	/* data fits into opal short atom */
 	if (src_len <= 15) {
-		if (len >= src_len + 1) {
```
AndrzejJakowski, line 164: please fix coding file for commenting and conditional block. for example:

/* comment 1*/
if (.....) {
    .....
/* comment 2*/
} else if (...) {
   .....
/* comment 3*/
} else {
   .....
}

src/lib/opal_parser.c
```diff
@@ -161,25 +181,28 @@ int append_bytes(uint8_t *buf, size_t len, const uint8_t *src, int src_len)
...
-		if (len >= src_len + 2) {
+		if (len >= src_len + 2)
 			ret += append_medium_atom_bytes_header(buf, len, src_len);
-			memcpy(buf + ret, src, src_len);
-			ret += src_len;
-		}
 	}
 
+	/* data fits into opal long atom */
+	else {
```
AndrzejJakowski: long also has limit to 16777215B. suggesting to put condition checking it. and assert if somebody provides larger buffer otherwise

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: The response we parse from the tper properties are of type uint64_t. Just to avoid data type mismatch, I'm using uint64_t type here for req(resp)_buf_size.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: That's true...I should have included a free(req_buf), in-case resp buf memory allocation failed.
But I like the idea you have suggested and will use it.

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: will do it

src/lib/opal_parser.c
RevanthRajashekar: will fix it

src/lib/opal_parser.c
RevanthRajashekar: will fix it

src/lib/opal_parser.c
RevanthRajashekar: Yeah I agree on that... will include the additional check

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: Also, instead of using sizeof(uint8_t), we can use sizeof(*dev->req_buf).

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: I'm ok to leave it as is, because I don't think it does any harm. It just caught my eye seeing 64B for buffer size

RevanthRajashekar:
Changes incorporated based on comments. Please review it.

src/lib/opal_parser.c
```diff
@@ -113,6 +113,26 @@ static int append_medium_atom_bytes_header(uint8_t *buf, size_t len, int data_le
...
 }
 
+static int append_long_atom_bytes_header(uint8_t *buf, size_t len, int data_len)
+{
+	uint8_t val = 0;
+
+	if (len >= 4) {
+		val = OPAL_ATOM_LONG_CODE;
+		val |= OPAL_ATOM_LONG_BYTESTRING;
+		val |= 0;
```
AndrzejJakowski: It looks like last instruction doesn't have any effect. suggest removing it

src/lib/opal_parser.c
```diff
@@ -160,26 +180,27 @@ int append_bytes(uint8_t *buf, size_t len, const uint8_t *src, int src_len)
...
 			ret += append_medium_atom_bytes_header(buf, len, src_len);
-			memcpy(buf + ret, src, src_len);
-			ret += src_len;
-		}
+
+	/* data fits into opal long atom */
+	} else if (src_len < 16777216) {
+		if (len >= src_len + 4)
+			ret += append_long_atom_bytes_header(buf, len, src_len);
 	}
```
AndrzejJakowski, line 202: please add assert for } else { buffer sizes

src/lib/nvme_pt_ioctl.c
```diff
@@ -361,13 +367,62 @@ void opal_deinit_pt(struct sed_device *dev)
...
+		free(dev->req_buf);
+
+	/*
+	 * Allocate memory for request and response buffers in a
+	 * single malloc request and split it later.
+	 */
+	ptr = malloc(sizeof(*dev->req_buf) * 2 * size);
+	if (!ptr)
+		return -ENOMEM;
+	memset(ptr, 0, sizeof(*ptr));
```
AndrzejJakowski: I think you intended to zero whole newly allocated buffer:
memset(ptr, 0, sizeof(*ptr) * 2 * size)

src/lib/nvme_pt_ioctl.c
```diff
@@ -400,14 +455,29 @@ int opal_init_pt(struct sed_device *dev, const char *device_path)
...
+	ret = resize_io_buf(opal_dev, MIN_IO_BUFFER_LEN);
 	if (ret)
 		goto init_deinit;
 
-	SEDCLI_DEBUG_PARAM("The device comid is: %u\n", opal_dev->comid);
+	ret = opal_dev_discv(dev);
+	if (ret)
+		goto init_deinit;
+
+	uint64_t max_com_pkt_sz = tper_prop_to_val(dev, MCPS);
```
AndrzejJakowski: please avoid mixing code with declaration of variable

src/lib/nvme_pt_ioctl.c
```diff
@@ -1565,11 +1645,12 @@ static int opal_generic_write_table(int fd, struct opal_device *dev,
...
 		 * The append_bytes used below, dependng upon the len either uses
 		 * short_atom_bytes_header (returns 1) or medium_atom_bytes_header
-		 * (returns 2). Hence we consider the MAX of the two i.e, 2.
+		 * (returns 2) or long_atom_bytes_header (returns 4).
+		 * Hence we consider the MAX of the three i.e, 4.
 		 *
 		 * The 1 byte is for the following ENDNAME token.
 		 */
-		uint64_t remaining_buff_size = buf_len - (pos + 2 + 1 + CMD_END_BYTES_NUM);
+		uint64_t remaining_buff_size = buf_len - (pos + 4 + 1 + CMD_END_BYTES_NUM);
```
AndrzejJakowski: please aovid mixing code with viariable declaration

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: My bad. will fix it

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: Okay will do it

src/lib/nvme_pt_ioctl.c
RevanthRajashekar: will fix it.

src/lib/opal_parser.c
RevanthRajashekar: This OR operation is to make sure that the sign/continued bit is set to zero in the Long atom header.
I would better keep this here.

src/lib/opal_parser.c
RevanthRajashekar: I was planning on introducing #defines for the max short, medium and long atom data lens.
Also, insert an assert at the beginning of the function rather than having it in an else condition.
assert (src_len < MAX_LONG_ATOM_DATA_LEN)
What do you think ?

src/lib/opal_parser.c
AndrzejJakowski: val is already zeroed in line 118

RevanthRajashekar: Ok will do it

RevanthRajashekar:
Changes included. Please review

PR #61, opened 2020-09-17
Only print discv supp

RevanthRajashekar:


RevanthRajashekar:
Will send out  a new pull request with new improvements.

PR #65, opened 2020-10-03
build: test compile whole sed_ioctl.c before turning on kernel opal d…

AndrzejJakowski:
…river

This patch introduces change to configure script to test compile whole
sed_ioctl.c file to determine if sedcli module using native kernel opal
driver should be used. This ensures that sed_ioctl.c will only be added
to sedcli binary if test compilation succeds and prevents situation
with kernel version incompatibilities.

Signed-off-by: Andrzej Jakowski <andrzej.jakowski@linux.intel.com>
---
Tested on:
 * Centos Linux 7.8.2003; 3.10.0-1127.8.2.el7.x86_64
 * CentOS Linux 8.2.2004; 4.18.0-193.el8.x86_64
 * Fedora 31; 5.5.5-200.fc31.x86_64
 * Ubuntu 19.04; 5.0.0-13-generic

### DIFF
--- a/src/configure
+++ b/src/configure
@@ -36,7 +36,13 @@ function print_status() {
 function test_compile() {
 	echo "Compiling test case $1" >> config.log
 
-	$cc $2 $3 -o $TMP_BIN $TMP_SRC >> config.log 2>&1 || return $?
+	if [ "$4" != "" ]; then
+		src=$4
+	else
+		src=$TMP_SRC
+	fi
+
+	$cc $2 $3 -o $TMP_BIN $src >> config.log 2>&1 || return $?
 }
 
 function generate_version() {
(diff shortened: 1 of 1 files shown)


WinsonYung:
I tried your patch on Ubuntu 20.04.1 LTS, when I run the following level0discovery cmd, I am still getting an error:

sudo ./sedcli -D --device /dev/nvme0n1
/dev/nvme0n1: Error initializing device

AndrzejJakowski:
> I tried your patch on Ubuntu 20.04.1 LTS, when I run the following level0discovery cmd, I am still getting an error:
> 
> sudo ./sedcli -D --device /dev/nvme0n1
> /dev/nvme0n1: Error initializing device

Failure you are seeing is not related to this patch as it fixes compilation problem on newer distributions e.g. RHEL/Centos 8.X.
Obviously sedcli compiled successfully and when run returns error. I would like to investigate that issue more

RevanthRajashekar:
Looks good to me

PR #66, opened 2020-10-04
libsed, sedcli: add Ruby discovery

AndrzejJakowski:
This patch add processing and displying of Ruby discovery header.

Signed-off-by: Andrzej Jakowski <andrzej.jakowski@linux.intel.com>


RevanthRajashekar:
Looks good to me.
Thanks for submitting the patch.

PR #67, opened 2020-10-04
libsed, sedcli: enable blockSID functionality per TCG BlockSID spec

WinsonYung:
This patch enables the support of BlockSID to block SID authority.
The feature supports using Clear Event flag that can determine the
blocking can be reset by SSD firmware during hardware reset.

Signed-off-by: Winson Yung <winson.yung@intel.com>

### DIFF
--- a/doc/sedcli.8
+++ b/doc/sedcli.8
@@ -91,6 +91,9 @@ MBR Done update to TRUE will only take effect when MBR Shadow is enabled.
 Write data into MBR shadow region(MBR table).
 MBR shadow should be enabled for the PBA to take effect.
 
+.IP "\fB\-B -d <device>\fR or \fB\-\-block_sid --device <device> --hwreset {1|0}\fR"
+Issue BlockSID authentication command with Clear Event flag via --hwreset option.
+
 .IP "\fB\-V\fR or \fB\-\-version\fR"
 Prints version of sedcli.
 
--- a/doc/sedcli.8
+++ b/doc/sedcli.8
@@ -91,7 +91,7 @@ MBR Done update to TRUE will only take effect when MBR Shadow is enabled.
 Write data into MBR shadow region(MBR table).
 MBR shadow should be enabled for the PBA to take effect.
 
-.IP "\fB\-B -d <device>\fR or \fB\-\-block_sid --device <device> --hwreset {1|0}\fR"
+.IP "\fB\-B -d <device> -r {1|0}\fR or \fB\-\-block_sid --device <device> --hwreset {1|0}\fR"
 Issue BlockSID authentication command with Clear Event flag via --hwreset option.
 
 .IP "\fB\-V\fR or \fB\-\-version\fR"
--- a/src/lib/nvme_access.c
+++ b/src/lib/nvme_access.c
@@ -44,11 +44,11 @@ static int send_recv_nvme_pt_ioctl(int fd, int send, uint8_t proto_id,
 	return status;
 }
 
-static int opal_send(int fd, uint16_t com_id, uint8_t *buf, int buf_len)
+int opal_send(int fd, uint8_t proto_id, uint16_t com_id, uint8_t *buf, int buf_len)
 {
 	int ret;
 
-	ret = send_recv_nvme_pt_ioctl(fd, SEND, TCG_SECP_01, com_id, buf,
+	ret = send_recv_nvme_pt_ioctl(fd, SEND, proto_id, com_id, buf,
 			buf_len);
 
 	return ret;
@@ -89,7 +89,7 @@ int opal_send_recv(int fd, uint16_t com_id, uint8_t *req_buf,
 {
 	int ret;
 
-	ret = opal_send(fd, com_id, req_buf, req_buf_len);
+	ret = opal_send(fd, TCG_SECP_01, com_id, req_buf, req_buf_len);
 	if (ret)
 		return ret;
 
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -40,6 +40,9 @@
 
 #define MIN_IO_BUFFER_LEN 2048
 
+#define OPAL_BLOCKSID_COMID  5
+#define BLOCKSID_PAYLOAD_SZ  512
+
 /* #define TPer property name strings here */
 #define MCPS "MaxComPacketSize"
 
@@ -2494,6 +2497,32 @@ int opal_write_shadow_mbr_pt(struct sed_device *dev, const struct sed_key *key,
 
 end_sessn:
 	opal_end_session(dev->fd, opal_dev);
+	return ret;
+}
+
+int opal_block_sid_pt(struct sed_device *dev, bool hw_reset)
+{
+	struct opal_device *device = dev->priv;
+	int ret, req_size = BLOCKSID_PAYLOAD_SZ;
+
+	/* Send Block SID authentication command, no IF_RECV response is expected */
+
+	if (device->req_buf_size < BLOCKSID_PAYLOAD_SZ) {
+		SEDCLI_DEBUG_MSG("Too small buffer size of BlockSID payload\n");
+		return -EINVAL;
+	}
+
+	/* Set Hardware Reset in LSB of the first byte in  */
+	/* BlockSID payload based on Clear Event flag user */
+	/* supplied through hwreset argument               */
+	*(device->req_buf) = hw_reset ? 1 : 0;
+
+	ret = opal_send(dev->fd, TCG_SECP_02, OPAL_BLOCKSID_COMID,
+							device->req_buf, req_size);
+	if (ret) {
+		SEDCLI_DEBUG_MSG("Error in NVMe passthrough ops\n");
+	}
+
 	return ret;
 }
 
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -2505,10 +2505,10 @@ int opal_block_sid_pt(struct sed_device *dev, bool hw_reset)
 	struct opal_device *device = dev->priv;
 	int ret;
 
-	/* Send Block SID authentication command, no IF_RECV response is expected */
-
-	/* Set Hardware Reset in LSB of the first byte in
-	 * BlockSID payload based on Clear Event flag user
+	/* Send Block SID authentication command, no
+	 * IF_RECV response is expected. Set Hardware
+	 * Reset in LSB of the first byte in BlockSID
+	 * payload based on Clear Event flag user
 	 * supplied through hwreset argument */
 
 	*(device->req_buf) = hw_reset ? 1 : 0;
--- a/src/lib/nvme_pt_ioctl.c
+++ b/src/lib/nvme_pt_ioctl.c
@@ -2503,22 +2503,18 @@ end_sessn:
 int opal_block_sid_pt(struct sed_device *dev, bool hw_reset)
 {
 	struct opal_device *device = dev->priv;
-	int ret, req_size = BLOCKSID_PAYLOAD_SZ;
+	int ret;
 
 	/* Send Block SID authentication command, no IF_RECV response is expected */
 
-	if (device->req_buf_size < BLOCKSID_PAYLOAD_SZ) {
-		SEDCLI_DEBUG_MSG("Too small buffer size of BlockSID payload\n");
-		return -EINVAL;
-	}
+	/* Set Hardware Reset in LSB of the first byte in
+	 * BlockSID payload based on Clear Event flag user
+	 * supplied through hwreset argument */
 
-	/* Set Hardware Reset in LSB of the first byte in  */
-	/* BlockSID payload based on Clear Event flag user */
-	/* supplied through hwreset argument               */
 	*(device->req_buf) = hw_reset ? 1 : 0;
 
 	ret = opal_send(dev->fd, TCG_SECP_02, OPAL_BLOCKSID_COMID,
-							device->req_buf, req_size);
+							device->req_buf, BLOCKSID_PAYLOAD_SZ);
 	if (ret) {
 		SEDCLI_DEBUG_MSG("Error in NVMe passthrough ops\n");
 	}
--- a/src/lib/sed.c
+++ b/src/lib/sed.c
@@ -47,6 +47,7 @@ typedef int (*ds_read)(struct sed_device *, enum SED_AUTHORITY, const struct sed
 typedef int (*ds_write)(struct sed_device *, enum SED_AUTHORITY, const struct sed_key *, const uint8_t *, uint32_t, uint32_t);
 typedef int (*list_lr)(struct sed_device *, const struct sed_key *,
 		       struct sed_opal_lockingranges *);
+typedef int (*blocksid)(struct sed_device *, bool);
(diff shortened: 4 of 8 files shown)


src/lib/sed.c
```diff
@@ -369,6 +373,11 @@ int sed_list_lr(struct sed_device *dev, const struct sed_key *key,
 	return curr_if->list_lr_fn(dev, key, lrs);
 }
 
+int sed_issue_blocksid_cmd(struct sed_device *dev, bool hw_reset)
+{
+	return curr_if->blocksid_fn(dev, hw_reset);
```
AndrzejJakowski, line 381: this needs to be check if blocksid_fn is not null -- otherwise it will crash. when blocksid_fn is null return -EOPNOTSUPP

src/sedcli_main.c
```diff
@@ -974,6 +1008,25 @@ static int handle_write_mbr(void)
...
+	struct sed_device *dev = NULL;
+	int ret = sed_init(&dev, opts->dev_path);
+
+	if (ret) {
+		sedcli_printf(LOG_ERR, "Error in initializing the dev: %s\n",
+				opts->dev_path);
+		return ret;
+	}
+
+	ret = sed_issue_blocksid_cmd(dev, opts->hardware_reset);
```
AndrzejJakowski, line 1026: do we need to make check if block sid feature is supported by tper before actually calling it?

src/lib/nvme_pt_ioctl.c
```diff
@@ -2497,3 +2500,29 @@ int opal_write_shadow_mbr_pt(struct sed_device *dev, const struct sed_key *key,
...
+	int ret, req_size = BLOCKSID_PAYLOAD_SZ;
+
+	/* Send Block SID authentication command, no IF_RECV response is expected */
+
+	if (device->req_buf_size < BLOCKSID_PAYLOAD_SZ) {
+		SEDCLI_DEBUG_MSG("Too small buffer size of BlockSID payload\n");
+		return -EINVAL;
+	}
+
+	// Set Hardware Reset in LSB of the first byte in
```
AndrzejJakowski: please use multi line commenting style: "/* */"

src/lib/nvme_pt_ioctl.c
```diff
@@ -2497,3 +2500,29 @@ int opal_write_shadow_mbr_pt(struct sed_device *dev, const struct sed_key *key,
...
 }
 
+int opal_block_sid_pt(struct sed_device *dev, bool hw_reset)
+{
+	struct opal_device *device = dev->priv;
+	int ret, req_size = BLOCKSID_PAYLOAD_SZ;
+
+	/* Send Block SID authentication command, no IF_RECV response is expected */
+
+	if (device->req_buf_size < BLOCKSID_PAYLOAD_SZ) {
```
AndrzejJakowski: req_buf_size -- is guaranteed to be at min 2048B in size. I don't think that this condition will ever be true

src/lib/nvme_pt_ioctl.c
```diff
@@ -40,6 +40,9 @@
 
 #define MIN_IO_BUFFER_LEN 2048
 
+#define OPAL_BLOCKSID_COMID  5
+#define BLOCKSID_PAYLOAD_SZ  512
```
AndrzejJakowski, line 44: is there a particular reason for this size? at minimum we support buffer 2048 B in size

src/lib/nvme_access.c
```diff
@@ -44,7 +44,7 @@ static int send_recv_nvme_pt_ioctl(int fd, int send, uint8_t proto_id,
 	return status;
 }
 
-static int opal_send(int fd, uint16_t com_id, uint8_t *buf, int buf_len)
```
AndrzejJakowski, line 47: add parameter to opal_send proto_id or somethig and make this function public. 
don't add opal_send_secp02

src/sedcli_main.c
```diff
@@ -101,6 +103,12 @@ static cli_option write_mbr_opts[] = {
 	{0}
 };
 
+static cli_option blocksid_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'r', "hwreset", "Clear events by setting hardware Reset flag. Allowed values: 1/0", 1, "FMT", CLI_OPTION_REQUIRED},
```
AndrzejJakowski: Spec has "hardware Reset" both capitilized. please stay consitent.
why not to have hwreset flag be optional with deafult setting "no". you can make it similar to "psid" flag in revert command

AndrzejJakowski:
Thx for the patch! please also update documentation (manpage  in doc directory) with description of new command

src/lib/nvme_pt_ioctl.c
WinsonYung: I don't see the blockSID spec mandate that, however I see our own tool is sending 512 byte to drive regardless the supported buffer size is.

src/lib/nvme_pt_ioctl.c
WinsonYung: I prefer to leave it there in case the underlying assumption changes.

src/lib/nvme_pt_ioctl.c
WinsonYung: sure

src/sedcli_main.c
WinsonYung: this is a tricky one that I think the tool shouldn't get too smart about that to make such decision for user. In other words, the user should check level0 discovery to see whether BlockSID feature descriptor is present in order to issue blockSID cmd. If drive doesn't support it, it should just drop or return error.

src/lib/sed.c
WinsonYung: I will add the check.

src/sedcli_main.c
WinsonYung: I will change hardware to Hardware for consistency. As to the optional flag, I personally like to be more expilict so user knows exactly what they supply when it comes to issue BlockSID.

src/lib/nvme_access.c
WinsonYung, line 47: I don't know what happen here, but I will remove the function name with secp at the end and passing in protocol id in instead.

src/sedcli_main.c
AndrzejJakowski: can you please explain why it is tricky?

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: I have just checked "Architecture Core spec" and it states in 5.2.2.4 that communication minimum is 1024B, instead of 2048B like I orginally mentioned. because of that this assumption will not change.

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: please take a look at architecture core specifiocation chapter 5.2.2.4 "Communications minimums"

src/sedcli_main.c
WinsonYung: I don't mean adding code to check block sid availability is tricky, but this can introduce issue such as drive may have bug in level0 discovery where it didn't expose block sid availability correctly. In this case, sedcli may incorrectly determine blocksid will not work on the drive, hence allowing user to checkout the feature.

src/lib/nvme_pt_ioctl.c
WinsonYung: Your reading of the core spec is incorrect, the minimum requirement under 5.2.2.4 suggests the minimum set of property name/value TPer and Host shall support, which defines MaxComPacketSize as 1024 for example. I double checked with the content expert on this. For BlockSID, you can in fact just have one byte, 512 byte, or 10224 bytes. They should be all fine.

src/lib/nvme_pt_ioctl.c
WinsonYung: see my other comment addressing your same concern.

src/lib/nvme_pt_ioctl.c
AndrzejJakowski: I didn't interpret spec incorrectly. I'm saying that req_buf_size is guaranteed in sedcli to be 1024 which corresponds to value from "communication mimimums" table.
Your if statement will always be false -- thus I want you to remove it.

src/lib/nvme_pt_ioctl.c
```diff
@@ -2497,3 +2500,25 @@ int opal_write_shadow_mbr_pt(struct sed_device *dev, const struct sed_key *key,
...
+{
+	struct opal_device *device = dev->priv;
+	int ret;
+
+	/* Send Block SID authentication command, no IF_RECV response is expected */
+
+	/* Set Hardware Reset in LSB of the first byte in  */
+	/* BlockSID payload based on Clear Event flag user */
+	/* supplied through hwreset argument               */
+
```
RevanthRajashekar, line 2513: How about merging these 4 different lines of  comments in standard way.
Something like this.   
/*
 *
 *
 */

src/sedcli_main.c
```diff
@@ -101,6 +103,12 @@ static cli_option write_mbr_opts[] = {
 	{0}
 };
 
+static cli_option blocksid_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'r', "hwreset", "Clear events by setting Hardware Reset flag. Allowed values: 1/0", 1, "FMT", CLI_OPTION_REQUIRED},
```
RevanthRajashekar, line 108: How about making it a Flag, rather than using it as a Format, taking values.

src/sedcli_main.c
```diff
@@ -404,6 +424,20 @@ int write_mbr_handle_opts(char *opt, char **arg)
...
 
+int blocksid_handle_opts(char *opt, char **arg)
+{
+	int hwreset_flag = 0; /* No reset BlockSID upon power events */
+
+	if (!strcmp(opt, "device")) {
+		strncpy(opts->dev_path, arg[0], PATH_MAX - 1);
+	} else if (!strcmp(opt, "hwreset")) {
+		hwreset_flag = atoi(arg[0]);
+		opts->hardware_reset = (hwreset_flag == 1) ? true : false;
```
RevanthRajashekar, line 439: This can be changed into 1 line after changing the hardware reset option to flag.
opts->hardware_reset = true;

src/sedcli_main.c
WinsonYung: Actually the intention is making this hwreset mandatory by specify a value. I would like to keep this way for user to specify a value explicitly rather than implicitly through the presence of the hwreset as a flag.

src/lib/nvme_pt_ioctl.c
WinsonYung: sure

src/sedcli_main.c
WinsonYung: See my previous comment, not intended to use this as a flag to indicate setting hardware reset or not. This is more of cosmetic change rather than functionality change which I would like to keep the current implementation.

doc/sedcli.8
```diff
@@ -91,6 +91,9 @@ MBR Done update to TRUE will only take effect when MBR Shadow is enabled.
 Write data into MBR shadow region(MBR table).
 MBR shadow should be enabled for the PBA to take effect.
 
+.IP "\fB\-B -d <device>\fR or \fB\-\-block_sid --device <device> --hwreset {1|0}\fR"
```
AndrzejJakowski: --hwreset option is missing in the long short form. Should be:
.IP "\fB\-B -d <device>-r {1|0} \fR or \fB\-\-block_sid --device <device> --hwreset {1|0}\fR"

src/sedcli_main.c
AndrzejJakowski: OK. If intention is to treat this option as integer accepting "0" or "1", please use appropriate flags (CLI_OPTION_RANGE_INT) and .min_value and max_value to enforce user input checks

src/sedcli_main.c
```diff
@@ -404,6 +424,20 @@ int write_mbr_handle_opts(char *opt, char **arg)
...
 }
 
+int blocksid_handle_opts(char *opt, char **arg)
+{
+	int hwreset_flag = 0; /* No reset BlockSID upon power events */
+
+	if (!strcmp(opt, "device")) {
+		strncpy(opts->dev_path, arg[0], PATH_MAX - 1);
+	} else if (!strcmp(opt, "hwreset")) {
+		hwreset_flag = atoi(arg[0]);
```
AndrzejJakowski, line 438: atoi will accept some random text e.g. "./sedcli -B -d /dev/nvme0n1 -r adsfdadaf" -- it needs to be changed

src/sedcli_main.c
```diff
@@ -404,6 +424,20 @@ int write_mbr_handle_opts(char *opt, char **arg)
 	return 0;
 }
 
+int blocksid_handle_opts(char *opt, char **arg)
+{
+	int hwreset_flag = 0; /* No reset BlockSID upon power events */
```
AndrzejJakowski: please don't append comment to code, rather put it line before

src/lib/nvme_pt_ioctl.c
```diff
@@ -2497,3 +2500,25 @@ int opal_write_shadow_mbr_pt(struct sed_device *dev, const struct sed_key *key,
 	return ret;
 }
 
+int opal_block_sid_pt(struct sed_device *dev, bool hw_reset)
+{
+	struct opal_device *device = dev->priv;
+	int ret;
+
+	/* Send Block SID authentication command, no IF_RECV response is expected */
```
AndrzejJakowski: why to have 2 comments one after another. Can you please merge it? e.g.
/* Send Block SID authentication command, no IF_RECV response is expected
 * Set Hardware Reset in LSB of the first byte in
 * BlockSID payload based on Clear Event flag user
 * supplied through hwreset argument */

AndrzejJakowski:
Hi, thx for the patch!
Please address inline comments

src/sedcli_main.c
RevanthRajashekar: You can use strtoul for a safer conversion.

src/sedcli_main.c
WinsonYung: sure

src/sedcli_main.c
WinsonYung: if I use CLI_OPTION_RANGE_INT with min and max, I don't believe I need to change here anymore, please confirm

doc/sedcli.8
WinsonYung: thanks

src/lib/nvme_pt_ioctl.c
WinsonYung: Done change

src/sedcli_main.c
AndrzejJakowski: random string is still passed to atoi() and accepted by app. I see 2 possibilities to fix it:
1. add appropriate checks in blocksid_handle_opts() and rejected invalid input
2. more general fix in args parse to prevent invalid inputs for INT
while 2 is more general and prefered approach; 1 is good tradeoff to introduce this feature quickly. I recommend 1

WinsonYung: Please check my updated code

PR #69, opened 2020-11-08
Multi usrs multi admin

RevanthRajashekar:


AndrzejJakowski (commented):
Can you please provide commit description describing introduced changes?

doc/sedcli.8
```diff
@@ -94,6 +94,18 @@ MBR shadow should be enabled for the PBA to take effect.
...
+.IP
+Add users to the Locking Ranges with the following access modes:
+.IP
+1. Read-Only(RO) - user can only read the data from the disk.
+.IP
+2. Read-Write(RW) - user can read/write data from/to the disk.
+.IP
+3. Locked(LK) - data is locked, user can NOT read/write data from/to the disk.
+.IP
+With a max. of 9 Locking Ranges available on a device, the value of N ranges from 1 to 9.
```
AndrzejJakowski, line 107: spec at minimum require 8 + 1 ranges. what if device supports more than 8+1 locking ranges?

src/sedcli_main.c
```diff
@@ -113,6 +115,14 @@ static cli_option blocksid_opts[] = {
 	{0}
 };
 
+static cli_option add_user_lr_opts[] = {
+	{'d', "device", "Device node e.g. /dev/nvme0n1", 1, "DEVICE", CLI_OPTION_REQUIRED},
+	{'u', "user", "String specifying the user to be added to the LR {userN, N=[1..9]}", 1, "FMT", CLI_OPTION_REQUIRED},
+	{'t', "accesstype", "String specifying access type to the data on drive. Allowed values: RO/RW/LK", 1, "FMT", CLI_OPTION_REQUIRED},
+	{'l', "locking-range", "Locking Range {1..9}", 1, "NUM", CLI_OPTION_OPTIONAL_ARG},
```
AndrzejJakowski, line 122: looks like "locking-range" is optional and you allow values from 1 to 9 as valid input. by default locking-range would be set to 0 which seems to be invalid.

AndrzejJakowski, line 122: what would be default value for this option?

AndrzejJakowski (commented):
There is no possibility to configure multiple locking ranges in sedcli but it is possible to associate user w/ it.
how is this supposed to work?

doc/sedcli.8
RevanthRajashekar: two options here.
1. do the level 0 discovery and use the SED Opal v2.00 features' Admin Authorities LSP Supported, User Authorities LSP Supported values to check the user provided values.
2. Ask the user to check those values via discover 0 command

src/sedcli_main.c
RevanthRajashekar: default value will be 0, which is the global locking ranges

doc/sedcli.8
```diff
@@ -94,6 +94,18 @@ MBR shadow should be enabled for the PBA to take effect.
 .IP "\fB\-B -d <device> -r {1|0}\fR or \fB\-\-block_sid --device <device> --hwreset {1|0}\fR"
 Issue BlockSID authentication command with Clear Event flag via --hwreset option.
 
+.IP "\fB\-U -d <device> -u <userN> -t <accesstype> [-l <locking-range>]\fR or \fB\-\-add-user-to-lr --device <device> --user <userN> --accesstype {RO|RW|LK} [--locking-range <locking-range>]\fR"
+.IP
+Add users to the Locking Ranges with the following access modes:
```
WinsonYung, line 99: I don't think there is such thing as each (user) authority has its own access mode. I need to confirm that though.

AndrzejJakowski: It is possible to set/clear: ReadLockEnabled, WriteLockEnabled, ReadLocked, WritedLocked and many other things int Locking Table. Given locking object (row in a Locking Table) can associated with user object. More details are in section 5.7.2.2. or TCG_storage_architecture_core spec...

WinsonYung:
Here is a summary of concerns from me for this PR:

1. when you create ACE for access to RL/RLE/WL/WLE, they are completely control separate access. In other words, RL/RLE are only for read, and WL/WLE are only for write, the add_user_to_lr() you asked me to look shows that if l_state == OPAL_RW, you construct ACE for write lock only.. There is no such thing as Read and Write. If you went to enable both read and write, you have to set two different ACE using two different UID. One for OPAL_LOCKINGRANGE_ACE_RDLOCKED_UID, and one for OPAL_LOCKINGRANGE_ACE_WRLOCKED_UID 
 
2. In this PR, what is the option '-t LK' intend to do? By the documentation, it seems to suggest LK means to let sedcli to lock userN issue read/write to a locking range. By default userN shouldn't have access to change RL/RLE/WL/WLE column values, and userN might be or might be not able to read/write to a locking range depends on how other authority (i.e. Admin) setup the locking range access control. 

3. The PR is misleading to let tool user thinking that each userN can have their own access control to a locking range, but these locking control can be overwritten by other authority such as Admin

PR #70, opened 2020-11-09
sedcli: Fix sed_err_to_text

RevanthRajashekar:
When the device returns Failed status 0x3f(63), the existing code
runs into a segmentation fault because its trying to access the
text OOB.
This patch fixes the issue and mitigates the segmentation fault.

### DIFF
--- a/src/lib/sed.c
+++ b/src/lib/sed.c
@@ -132,32 +132,27 @@ static struct opal_interface nvmept_if = {
 
 static struct opal_interface *curr_if = &nvmept_if;
 
-struct sed_status_ret {
-	int code;
-	const char *text;
-};
-
-struct sed_status_ret sed_statuses[] = {
-	{ .code = SED_SUCCESS, .text = "Success" },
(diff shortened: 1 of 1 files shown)


src/lib/sed.c
```diff
@@ -387,5 +382,5 @@ const char *sed_error_text(int sed_status)
 		return NULL;
```
AndrzejJakowski, line 382: with changes you introduced I think this condition could be updated in following way:
if (sed_status < 0 || sed_status > SED_FAIL) {
    return NULL;
}
for non-existing elements between SED_AUTHORITY_LOCKED_OUT and SED_FAIL null should be returned as this elements are not defined.
when new status codes will be added this function will automatically work.

RevanthRajashekar: I agree with the 1st check condition. For non-existing elements between SED_AUTHORITY_LOCKED_OUT and SED_FAIL, the array has it initialized it to NULL, so we dont have to worry about it.
Yes! any new elements in between will automatically work :)

AndrzejJakowski:
looks good to me! thx!

PR #72, opened 2021-01-06
Rename and clarify enum opal_lock_state definition to align with TCG spec

WinsonYung:


WinsonYung:
In sedcli, there are two functions that hooked into .lock_unlock_fn where used to alter a locking range read/write permission. 

First function is sedopal_lock_unlock() which passes these locking range attribute (i.e. authority and locking type) into linux kernel through IOCTL to let kernel driver handles the locking range permission alteration, which works based on how linux kernel implemented and by how sedcli exposed through -accesstype. In this case kernel drive support these three type of accesses (Read-Only - WL/WLE = 1, RL/RLE = 0, Read & Write - WL/WLE = 0, RL/RLE = 0, or NoAccess - WL/WLE = 1, RL/RLE = 1) through IOCTL.

However in the second function opal_lock_unlock_pt() which uses opal_lock_unlock_no_sum() to only alter RL and WL (not RLE and/or WLE), and construct a tokenized payload to a Opal enabled SSD. In this case, it is not guaranteed that a locking range will have the correct access permission because RLE and WLE are left without being configured if secli is not used to setup the locking range initially. To have a locking range read-only permission, both RL and RLE need to be 1, and WL and WLE need to be 0. To have a locking range read & write permission, RL, RLE, WL and WLE all needs to be 1. To have a locking range NoAccess, RL, RLE, WL and WLE all needs to be 0. There is possibility that WLE and/or RLE are 1, and upon reset/power cycle, SSD firmware will turn RL and/or WL to 1 due LockOnReset behavior.

src/lib/opal_parser.h
```diff
@@ -66,10 +66,17 @@ enum opal_user {
...
 };
 
+/* The following enum are mapped to
+   Opal read and write locking state,
+   OPAL_RO requires both RL & RLE on a
+   locking range to be enabled in order
+   to be effective. OPAL_WR requires
+   both WL & WLE on a locking range to
+   be enabled in order to be effective */
+
```
RevanthRajashekar, line 76: The description here is really helpful.
Why are we removing the OPAL_LK here? This is required to lock the device from either reading or writing.
By RW, we meant to let the users Read/Write  from/to the device. 
I don't think it makes sense to let the users only write to the device while the read lock is enabled and can't read from the device.

WinsonYung: @RevanthRajashekar, please have a look of my other general comment in the PR. I can see the current enum definition for sedopal_lock_unlock (which i think is used when sedcli is using with opal driver) make sense. However with non-opal driver (i.e NVMe PT), we sending the tokenized payload to SSD dirctly, there is no direct mapping for OPAL_RW and OPAL_LK to a locking range state defined by Locking SP table in the spec. In current implementation, you are only changing RL/WL, how do you ensure RLE and WLE are set correctly?

WinsonYung: Maybe enum opal_lock_state is *not* intended to map to Locking SP RL/RLE/WL/WLE after all.

RevanthRajashekar: The --lock-unlock  functionality is used to Lock or unlock global locking range.
Before using it, make sure that you have setup the global locking range using the --setup-global-range command, which enables those RLE and WLE columns.

PR #73, opened 2021-02-20
Fix incorrect endianness with CNL level0 descriptor

utpalsolanki:
The Max Key and Unused Key fields are encoded in little endian they should be encoded in big endian.

TCG Storage Spec v2.01:

3.2.2.2 Endianness
The endianness of integers transmitted across the interface is big endian


utpalsolanki:
@RevanthRajashekar @WinsonYung @AndrzejJakowski @markbrown314 please review this pull-request. Thanks.

RevanthRajashekar (approved):
Looks good to me
Thanks for the patch

PR #78, opened 2021-07-09
libsed: opal_parser: Put token back to free list when error occurs

RevanthRajashekar:
In the fucntion opal_get_next_token, we are neglecting the token allocated
when there is an error parsing or the response payload has ended.
As a result we keep missing out a token while deallocating tokens and
putting them back to the list of free tokens -> number of available tokens
keeps reducing every time by 1.
This patch fixes the issue.


AndrzejJakowski:
This is great catch! Merging...

PR #80, opened 2021-08-14
Allow setting MBR done while keeping setting or MBR shadow

gzxu:
`sedcli -M -d /dev/sda -m TRUE` always gives `Error: disabling MBR shadow and setting MBR done doesn't take any effect`, this is not expected


RevanthRajashekar (approved):

RevanthRajashekar:
Thanks for the patch.

PR #83, opened 2022-02-01
Miscellaneous cosmetic commits

dougmill-ibm:
These commits address things found while preparing for larger changes to support abstracting the acquisition of the drive password.

### DIFF
--- a/src/sedcli_main.c
+++ b/src/sedcli_main.c
@@ -471,13 +471,15 @@ static void print_sed_status(int status)
 			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
 		} else if (status == -ENOMEM) {
 			sedcli_printf(LOG_ERR, "No memory\n");
+		} else if (status == -EOPNOTSUPP) {
+			sedcli_printf(LOG_ERR, "Not supported\n");
 		} else {
-			sedcli_printf(LOG_ERR, "Unknown error\n");
+			sedcli_printf(LOG_ERR, "Unknown error: %d\n", status);
 		}
 	} else {
 		sed_status = sed_error_text(status);
 		if (sed_status == NULL)
-			sedcli_printf(LOG_ERR, "Unknown Error\n");
+			sedcli_printf(LOG_ERR, "Unknown Error: %d\n", status);
 		else
 			sedcli_printf((status == SED_SUCCESS) ? LOG_INFO : LOG_ERR,
 					"%s\n", sed_status);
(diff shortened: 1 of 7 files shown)


src/sedcli_util.c
```diff
@@ -111,6 +112,7 @@ void *alloc_locked_buffer(size_t size)
 	if (!buf)
 		return NULL;
 
+	memset(buf, 0, 1);	/* silence GCC warning (error) */
```
gjoyce-lifesize, line 115: Any reason not to zero all of buf?

src/sedcli_main.c
```diff
@@ -471,13 +471,15 @@ static void print_sed_status(int status)
 			sedcli_printf(LOG_ERR, "Couldn't determine device state\n");
 		} else if (status == -ENOMEM) {
 			sedcli_printf(LOG_ERR, "No memory\n");
+		} else if (status == -EOPNOTSUPP) {
+			sedcli_printf(LOG_ERR, "Not supported\n");
 		} else {
-			sedcli_printf(LOG_ERR, "Unknown error\n");
+			sedcli_printf(LOG_ERR, "Unknown error: %d\n", status);
```
gjoyce-lifesize, line 477: I like printing status when it is unknown.

gjoyce-lifesize (approved):
I like the changes to use sed_key as a parameter since those functions just built a sed_key to use anyway.

src/sedcli_util.c
dougmill-ibm: Not really, other than just to limit the amount of work done. I wanted to just do "*buf = 0;" but 'buf' is 'void *' and the casting necessary was not pretty. This path is not performance critical, so could just zero all of it. I was annoyed that I had to touch *buf at all, but trying to fix this in GCC seems excessive. It will be a problem for our target distribution, though, as GCC 11 is already in use there. I didn't try it, but I suspect GCC would also be fooled by doing a memset of zero bytes.

dougmill-ibm: Perhaps, rather than use memset, the line should be:
```
*(char *)buf = 0;	/* silence GCC warning (error) */
```
It also turns out that the #include <string.h> is not necessary, perhaps due to side-effects of other #includes. 'memset' (et al.) is already used in the source file, and there are no warnings about it.

PiotrRudnickiAtSolidigm:
Hi,
this project is being continued,
please check sedcli new repository: https://github.com/solidigm/sedcli
There will be a massive update soon!

